PR #3, opened 2023-03-29
Create django app

SteffiStuffel:


integreat_compass/asgi.py
timoludwig, line 1: If we use the same deployments as for our other Django-systems, we don't need this file and just go with `wsgi.py`.

integreat_compass/settings.py
```diff
@@ -0,0 +1,115 @@
...
+
+# Build paths inside the project like this: BASE_DIR / 'subdir'.
+BASE_DIR = Path(__file__).resolve().parent.parent
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/4.1/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = "django-insecure-cy1_$eh*&b&@t9^13^*#el48a(^n^=(%s%shbpzyxe3y49qh+("
```
timoludwig, line 23:
```suggestion
SECRET_KEY = os.environ.get("INTEGREAT_CMS_SECRET_KEY", "dummy" if DEBUG else "")
```
I like the security measure we have for integreat-cms here... this prevents us from accidentally deploying this with the insecure dummy key, because an empty key will immediately raise an error if the correct key is not passed via the environment.

integreat_compass/settings.py
```diff
@@ -0,0 +1,115 @@
...
+
+Generated by 'django-admin startproject' using Django 4.1.2.
+
+For more information on this file, see
+https://docs.djangoproject.com/en/4.1/topics/settings/
+
+For the full list of settings and their values, see
+https://docs.djangoproject.com/en/4.1/ref/settings/
+"""
+
```
timoludwig, line 12:
```suggestion
import os
from distutils.util import strtobool
```

integreat_compass/settings.py
```diff
@@ -0,0 +1,115 @@
...
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = True
+
+ALLOWED_HOSTS = []
+
+
+# Application definition
+
+INSTALLED_APPS = [
+    "integreat_compass.cms",
```
timoludwig, line 34:
```suggestion
```
As long as the `cms` app does not exist, we cannot install it.

integreat_compass/settings.py
```diff
@@ -0,0 +1,115 @@
...
+    {"NAME": "django.contrib.auth.password_validation.MinimumLengthValidator"},
+    {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"},
+    {"NAME": "django.contrib.auth.password_validation.NumericPasswordValidator"},
+]
+
+
+# Internationalization
+# https://docs.djangoproject.com/en/4.1/topics/i18n/
+
+LANGUAGE_CODE = "en-us"
```
timoludwig, line 98:
```suggestion
LANGUAGE_CODE = "de"
```
I think we probably want German to be the default UI language?

manage.py
timoludwig, line 1: If we want to follow the same patterns as our other Django CMS systems, we could rename this file to `integreat-compass-cli`

timoludwig (changes requested):
Thanks a lot! :+1:

Looks like a pretty good skeleton. Just a few minor comments.

integreat-compass-cli.py
timoludwig, line 1: This file needs to be part of the python package directory `integreat_compass`.
Also, if we want to follow the same naming conventions as the other projects, we could remove the `.py` extension. This means it cannot be imported that easily, but it is meant to be executed directly rather than imported as python module, so I guess that's fine.

integreat-compass-cli.py
```diff
@@ -0,0 +1,27 @@
...
+import os
+import sys
+
+
+def main():
+    """Run administrative tasks."""
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "integreat_compass.settings")
+
+    # Override default port for `runserver` command
+    from django.core.management.commands.runserver import Command as runserver
```
timoludwig, line 12: Is there anything that speaks against moving this up to the regular imports?

timoludwig (changes requested):

integreat-compass-cli.py
SteffiStuffel: Yes did this in the next ticket, i am already working on. Will upload soon.

integreat-compass-cli.py
SteffiStuffel: No, will do.

timoludwig (approved):
:tada:

PR #12, opened 2023-03-29
Add `pyproject.toml`

timoludwig:
The tooling around it might be a bit over-engineered at the moment, but it could be useful in the future.
If we don't care about all the requirements checks, the install script could also be replaced by:
```
source .venv/bin/activate
pip install -e .[dev-pinned,pinned]
```


charludo:
> The tooling around it might be a bit over-engineered at the moment, but it could be useful in the future. If we don't care about all the requirements checks, the install script could also be replaced by:
> 
> ```
> source .venv/bin/activate
> pip install -e .[dev-pinned,pinned]
> ```

Overengineered or not, you've already spent the time for this on Integreat, so no point in not using it IMO 😄

SteffiStuffel (commented):
for consistency with integreat-cms, if would be nice if you can rename the "tools" folder to "dev-tools"

timoludwig:
> for consistency with integreat-cms, if would be nice if you can rename the "tools" folder to "dev-tools"

Since we also use `tools` in all other digitalfabrik projects (also lunes cms & app and integreat app), I renamed the directory in integreat-cms instead, see https://github.com/digitalfabrik/integreat-cms/commit/0fe94dda96a2dbd5feec07195259bd855cbccaad

timoludwig:
@SteffiStuffel and `sudo apt install python3.9-venv` does not fix the issue? :thinking: :cry:

SteffiStuffel:
> 

Yes it fixes it, but this step is not needed for integreat-cms

Issue #16, opened 2023-03-30
Add prettier formatter


charludo:
We should conform to one code style right from the beginning, not just for python code (#9) but also for JS/TS/CSS/HTML.

- Add prettier as dev dependency
- copy over (and possibly adjust) the [config](https://github.com/digitalfabrik/integreat-cms/blob/develop/.prettierrc) and [ignore](https://github.com/digitalfabrik/integreat-cms/blob/develop/.prettierignore) files from Integreat
- Add CircleCI job to check formatting

PR #20, opened 2023-04-12
Setup development tools

SteffiStuffel:
### Short description
Add run script and all necessary stuff around it. Splitted `_functions.sh` file into multiple files. Ports for dev-server and db are different from the ones used in integreat.

### Resolved issues

Fixes: #5

PR #21, opened 2023-04-17
Add prettier config

charludo:
### Short description

Add prettier config

### Proposed changes

- mostly just copied over from `integreat-cms`
- some small tweaks to `.prettierrc` that now closer resemble the defauls recommended in the docs 

### Side effects

- none

### Resolved issues

Fixes: #16


.circleci/config.yml
```diff
@@ -30,6 +30,47 @@ jobs:
...
             - integreat_compass.egg-info
+  npm-install:
+    docker:
+      - image: "cimg/node:lts"
+    resource_class: small
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - npm-{{ checksum "package-lock.json" }}-v2
```
timoludwig, line 41:
```suggestion
            - npm-{{ checksum "package-lock.json" }}-v1
```
Starting with v1 is probably fine :sweat_smile:

.circleci/config.yml
```diff
@@ -30,6 +30,47 @@ jobs:
...
+            else
+              npm ci
+              if [[ -n $(git diff --shortstat package-lock.json) ]]; then
+                echo "package-lock.json is out of date:"
+                git --no-pager diff package-lock.json
+                exit 1
+              fi
+            fi
+      - save_cache:
+          key: npm-{{ checksum "package-lock.json" }}-v2
```
timoludwig, line 56:
```suggestion
          key: npm-{{ checksum "package-lock.json" }}-v1
```

.circleci/config.yml
```diff
@@ -30,6 +30,47 @@ jobs:
...
+          command: |
+            if [[ -d "node_modules" ]]; then
+              echo "Node modules restored from cache, skipping npm install"
+            else
+              npm ci
+              if [[ -n $(git diff --shortstat package-lock.json) ]]; then
+                echo "package-lock.json is out of date:"
+                git --no-pager diff package-lock.json
+                exit 1
+              fi
```
timoludwig, line 53:
```suggestion
```
This does not make sense - `npm ci` does not rewrite the the lock file, so it will never be out of date (and on the other hand: if we use `npm install` here, it would not indicate a problem, since the generated lock file depends on the system it's generated on and is not a single source of truth)...
Probably, we should change this in the integreat-cms as well...

tools/prettier.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# This script can be used to format the static files with prettier
+
+# Import utility functions
+# shellcheck source=./tools/_functions.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_functions.sh"
```
timoludwig, line 7:
```suggestion
# shellcheck source=./tools/utils/_functions.sh
source "$(dirname "${BASH_SOURCE[0]}")/utils/_functions.sh"
```
The location will likely be renamed in #20

timoludwig (changes requested):
Thanks a lot!

What do you think about adding a pre-commit hook as well?

.circleci/config.yml
charludo: Oh, whoops :D

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "devDependencies": {
+    "prettier": "^2.8.4"
```
timoludwig, line 3:
```suggestion
    "prettier": "^2.7.1"
```
Apparently, the pre-commit mirror is a bit behind - but probably it's preferable to have the same versions in CircleCI and in the pre-commit hook?

package-lock.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "integreat-compass",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "devDependencies": {
+        "prettier": "^2.8.4"
```
timoludwig, line 8: Could you run `npm install` one last time to update the version here as well?

timoludwig (changes requested):

package-lock.json
charludo: Sorry about all those details :(

timoludwig (approved):
:tada:

Issue #17, opened 2023-03-30
Add eslint


charludo:
To stay consistent and prevent some easily avoidable problems, we should lint our JS/TS files right from the get-go.

- copy over (and possibly modify) the [config](https://github.com/digitalfabrik/integreat-cms/blob/develop/.eslintrc.js) from Integreat
- required packages (for the exact config in Integreat) that should be added as devDependencies (we might not need all of them) are:
    - "@typescript-eslint/eslint-plugin"
    - "eslint"
    - "eslint-config-airbnb"
    - "eslint-config-preact"
    - "eslint-plugin-import"
    - "eslint-plugin-jest"
    - "eslint-plugin-jsx-a11y"
    - "eslint-plugin-prefer-arrow"
    - "eslint-plugin-prettier"
    - "eslint-plugin-react"
    - "eslint-plugin-react-hooks"
    - "eslint-plugin-styled-components-a11y" 
- Add CircleCI job to test code against the linter

PR #22, opened 2023-04-17
Add eslint

charludo:
### Short description

Add eslint as linter for JS/TS files

### Proposed changes

- largely copied over eslint config, scripts, and dependencies from `integreat-cms`
- these are actually a little more strict than the version in `integreat-cms`, namely the following are enabled here:
    - no explicit any
    - explicit return types from modules
    - no circular imports
    - strict Null checks

### Side effects

- to not re-do what has already been done in #21, I've ommitted the `npm-install` hook in the CircleCI config

### Resolved issues

Fixes: #17

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,6 +55,17 @@ jobs:
           root: .
           paths:
             - node_modules
+  eslint:
+    docker:
+      - image: "cimg/node:lts"
+    resource_class: small
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Lint static CSS, JS & YAML files
+          command: npx eslint . --no-error-on-unmatched-pattern
   prettier:
     docker:
       - image: "cimg/node:lts"
@@ -78,6 +89,9 @@ workflows:
           filters:
             branches:
               ignore: main
+      - eslint:
+          requires:
+            - npm-install
       - prettier:
           requires:
             - npm-install
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,134 @@
+module.exports = {
+    env: {
+        es6: true,
+        browser: true,
+    },
+    ignorePatterns: [
+        "**/dist/",
+        "**/htmlcov/**",
+        "**/node_modules/",
+        "*.config.js",
+        "*.json",
+        "*.lock",
+        ".eslintrc.js",
+        ".venv/",
+        "build/",
+        "docs/",
+        "package.json",
+        "package-lock.json",
+    ],
+
+    extends: [
+        "airbnb",
+        "airbnb/hooks",
+        "eslint-config-preact",
+        "plugin:@typescript-eslint/recommended",
+        "plugin:import/typescript",
+    ],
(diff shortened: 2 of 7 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,137 @@
...
+    "react/jsx-indent-props": "off",
+    "react/jsx-one-expression-per-line": "off",
+    "wrap-iife": "off",
+
+    // overly strict rules
+    "@typescript-eslint/strict-boolean-expressions": "off",
+    "function-paren-newline": "off",
+    "jsx-a11y/label-has-associated-control": "off",
+    "jsx-a11y/no-noninteractive-element-interactions": "off",
+    "jsx-a11y/no-static-element-interactions": "off",
```
charludo, line 61: I wasn't sure if we'd maybe want re-enable these? They enforce some accessibility standards, e.g. not attaching (user-)event listeners to static elements like `div`s unless we have given that element an `aria-role` of (for example) `button`.
I disabled them in `integreat-cms` because these rules were broken a *lot*.

.circleci/config.yml
```diff
@@ -55,6 +55,17 @@ jobs:
...
+    docker:
+      - image: "cimg/node:lts"
+    resource_class: small
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Lint static CSS, JS & YAML files
+          command: npx eslint . --no-error-on-unmatched-pattern
```
charludo, line 68: The flag should probably be removed once the first actual JS/TS file is added, but without the CircleCI checks for this branch will not pass 😅

.pre-commit-config.yaml
```diff
@@ -1,4 +1,15 @@
 repos:
+  - repo: https://github.com/pre-commit/mirrors-eslint
+    rev: v8.27.0
```
timoludwig, line 3: Same discussion as in #21: Do we want the same version here as in the pckage-lock.json?
At least the eslint mirror seems to be a bit more up to date, so version `8.38.0` is the latest version in both repositories.

.eslintrc.js
timoludwig: I don't have a strong opinion here, but I guess it would be fine to start with a very strict config and only if you notice you spend a lot of your developing time fixing debatable warnings, we could disable them again...

timoludwig (approved):
Thanks! :+1:

.pre-commit-config.yaml
timoludwig: Ah, the `^8.27.0` in the package.json means "compatible with `8.27.0`", thus it installed `8.38.0` in the lock file (it's the same for prettier - `^2.7.1` still installs the newest available version [`2.8.7`](https://github.com/digitalfabrik/integreat-compass/blob/f016d20595bd9b365bf28fbbd3440f99c292db65/package-lock.json#L12)).
But ok, since they are compatible, a few minor versions difference shouldn't cause a different output...

timoludwig: Then I guess it can stay as it is :see_no_evil:

PR #24, opened 2023-04-19
Setup Testing

### LINKED ISSUE #6: Setup testing framework
Setup basic testing framework, e.g. pytest.
Structure and many utilities can be copied from https://github.com/digitalfabrik/integreat-cms/tree/develop/tests

recommended pytest extensions:
-    "pytest-circleci-parallelized"
-    "pytest-cov"
-    "pytest-django"
-    "pytest-testmon"
-    "pytest-xdist"
-    "pytest-icdiff"

ztefanie:
### Short description
Added testing config 

### Side effects
Changed SCRIPT_PATH calculation, because it otherwise looked for the script at the wrong place. Tested the other script, should work as expected, but it is not reall nice to hard code `tools` folder name here. Any better ideas?

### Resolved issues

Fixes: #6


tools/utils/_functions.sh
```diff
@@ -34,7 +34,7 @@ PACKAGE_DIR="${BASE_DIR}/${PACKAGE_DIR_REL}"
 # The filename of the currently running script
 SCRIPT_NAME=$(basename "$0")
 # The absolute path to the currently running script (required to allow restarting with different permissions
-SCRIPT_PATH="${DEV_TOOL_DIR}/${SCRIPT_NAME}"
+SCRIPT_PATH="${BASE_DIR[0]}/tools/${SCRIPT_NAME}"
```
timobrembeck: > Changed SCRIPT_PATH calculation, because it otherwise looked for the script at the wrong place. Tested the other script, should work as expected, but it is not reall nice to hard code tools folder name here. Any better ideas?

Maybe change the variables above?
```
DEV_TOOL_DIR=$(pwd)
cd ../..
```
to
```
cd ..
DEV_TOOL_DIR=$(pwd)
cd ..
```
or something similar? Then we can rely on the fact that `DEV_TOOL_DIR` is really the dev tool dir, in case we copy more dev tools from integrat-cms...

and keep this  line as it was?

```suggestion
SCRIPT_PATH="${DEV_TOOL_DIR}/${SCRIPT_NAME}"
```

tests/cms/test_dummy.py
```diff
@@ -0,0 +1,8 @@
+import json
+
+from integreat_compass.cms.models.justfortesting import add
+
+def test_dummy():
+    with open("integreat_compass/cms/fixtures/test_data.json", encoding="utf-8") as f:
+        result = json.load(f)
+    assert result == []
```
timobrembeck:
```suggestion

def test_dummy(load_test_data):
    pass

```
Maybe this makes more sense as dummy test? This also checks whether the test data has been loaded via the fixture.

tests/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+from django.core.management import call_command
+
+#: A role identifier for superusers
+ROOT = "ROOT"
+#: A role identifier for anonymous users
+ANONYMOUS = "ANONYMOUS"
+
+#: Enable the aiohttp pytest plugin to make use of the test server
+# pytest_plugins = "aiohttp.pytest_plugin"
+
```
```suggestion
```
Can be safely removed I guess. Aiohttp is probably not going to be needed in this project.

timobrembeck (changes requested):

pyproject.toml
```diff
@@ -47,6 +53,13 @@ dev-pinned = [
     "packaging==23.0",
     "pyproject_hooks==1.0.0",
     "tomli==2.0.1",
+    "pytest==7.3.1",
+    "pytest-circleci-parallelized==0.1.0",
+    "pytest-cov==4.0.0",
+    "pytest-django==4.5.2",
+    "pytest-testmon==1.4.5",
+    "pytest-xdist==3.2.1",
+    "pytest-icdiff==0.6",
```
timobrembeck: These package have quite a few sub-dependencies, so just listing the direct pinned versions is not the desired workflow. Could you instead calculate all installed packages recursively by using my script `./tools/update_dependencies.sh` (after rebasing)?

tests/cms/test_dummy.py
```diff
@@ -0,0 +1,2 @@
+def test_dummy(load_test_data):
+    pass
```
timobrembeck:
```suggestion
def test_dummy(load_test_data):
    pass

```
I think newlines at end of files should always be present

timobrembeck (changes requested):

pyproject.toml
ztefanie: Thanks, I did this. Can you please explain (or share a link) why this is done?

timobrembeck: For the reproducibility of the dev and prod environments.
Typically, python libraries don't pin their versions themselves, but just give a version range they're compatible with, to make sure they don't cause version conflicts when used together with other libraries.
So if we only pin the versions of the direct dependencies, the sub-dependencies (and sub-sub etc.) might still have different versions, causing e.g. different behaviours, incompatibilities or breaking changes that are only present on some systems, making it hard to debug the problem and determining its origin.
When pinning the dependencies recursively on the other hand, we can be 100% sure that we have the same packages installed locally, on CircleCI and finally on the production system.

timobrembeck (approved):
Thanks a lot! :muscle:

PR #26, opened 2023-04-19
Add black formatter

### LINKED ISSUE #9: Add black formatter
I we want to use the same guidelines as lunes-cms and integreat-cms, we could use black as formatter for this project.
- Add black as dev dependency
- black doesn't allow a lot of config, but I suggest e.g. `skip-magic-trailing-comma = true`
- Add CircleCI job to check formatting

JoeyStk:
### Short description
This PR adds the black formatter

### Proposed changes

- Add black formatter to pyproject.toml
- Add black as tool
- Add black as pre-commit
- Exclude skip-magic-trailing-comma as suggested in issue description

### Side effects

- I hope none

### Resolved issues

Fixes: #9

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,3 +14,7 @@ repos:
     rev: v2.7.1
     hooks:
       - id: prettier
+  - repo: https://github.com/psf/black
+    rev: 23.3.0
+    hooks:
+      - id: black
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ dynamic = ["version"]
 
 [project.optional-dependencies]
 dev = [
+    "black",
     "build",
     "pylint",
     "pylint-django",
@@ -61,6 +62,13 @@ dev-pinned = [
     "pyproject_hooks==1.0.0",
     "tomlkit==0.11.7",
     "wrapt==1.15.0",
+    "black==23.3.0",
+    "click==8.1.3",
+    "mypy-extensions==1.0.0",
+    "packaging==23.1",
+    "pathspec==0.11.1",
+    "platformdirs==3.2.0",
+    "pyproject_hooks==1.0.0",
 ]
 
 [project.urls]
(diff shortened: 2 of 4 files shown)


tools/black.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# Import utility functions
+# shellcheck source=./tools/utils/_functions.sh
+source "$(dirname "${BASH_SOURCE[0]}")/utils/_functions.sh"
+
+require_installed
+
+# Run black
+echo "Starting code formatting with black..." | print_info
+black "${BASE_DIR}"
```
timobrembeck:
```suggestion
black .
```
I think the current dir is the base dir anyway (since `_functions.sh` changes into the base dir)

tools/black.sh
```diff
@@ -0,0 +1,14 @@
...
+echo "✔ Code formatting finished" | print_success
```
timobrembeck:
```suggestion
echo "✔ Code formatting finished" | print_success

```

.pre-commit-config.yaml
```diff
@@ -14,3 +14,7 @@ repos:
     rev: v2.7.1
     hooks:
       - id: prettier
+  - repo: https://github.com/psf/black
+    rev: 23.1.0
```
timobrembeck:
```suggestion
    rev: 23.3.0
```
I think it makes sense to use the same version like in the pyproject.toml

timobrembeck (changes requested):

tools/black.sh
JoeyStk: Yes, of course

timobrembeck (approved):

pyproject.toml
```diff
@@ -128,3 +136,6 @@ enable = [
 
 [tool.pylint.reports]
 output-format = "colorized"
+
+[tool.black]
+skip-magic-trailing-comma = true
```
timobrembeck: We could sort this alphabetically, so `[tool.black]` before `[tool.pylint.main]`?

timobrembeck (changes requested):

timobrembeck (approved):

PR #27, opened 2023-04-19
Add default pre-commit hooks

timobrembeck:
### Short description

Enable a few default pre-commit hooks which sanitize the general file format and try to rule out some common mistakes


ztefanie (approved):
Looks good, but maybe it makes sense to wait for the isort and djlint PRs, so they can directly be added.

timobrembeck:
> Looks good, but maybe it makes sense to wait for the isort and djlint PRs, so they can directly be added.

Since #32 and #28 both have conflicts with the base branch right now, I guess it's fine to merge this one now...

PR #30, opened 2023-04-19
Improve dependency update script

timobrembeck:
### Short description

Improve dependency update script

### Proposed changes

- Add the possibility to autoupdate pre-commit hook versions
- Add the possibility to only update one lock file by passing the `--only` flag
- Call npm update with the `--save` flag to also update the `package.json` and allow updating major versions

### DIFF
--- a/tools/update_dependencies.sh
+++ b/tools/update_dependencies.sh
@@ -14,49 +14,89 @@ function format_pyproject_toml {
     done
 }
 
+function update_npm_dependencies {
+    # Check if npm dependencies are up to date
+    echo "Updating JavaScript dependencies..." | print_info
+    npm update --save
+
+    # Fix npm security issues (skip all breaking changes)
+    echo "Running security audit of JavaScript dependencies..." | print_info
+    npm audit fix || true
+}
+
+function update_pip_dependencies {
+    # Update pip dependencies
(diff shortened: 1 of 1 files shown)


tools/update_dependencies.sh
```diff
@@ -14,49 +14,89 @@ function format_pyproject_toml {
...
+        --expression "/^dev-pinned = \[$/,/^\]$/c\dev-pinned = [\n${PINNED_DEV_VERSIONS_TOML}]" \
+        pyproject.toml
+    # Remove the temporary venv
+    deactivate
+    rm -rf .venv.tmp
+
+    # Install updated versions in the real venv
+    if [[ -d ".venv" ]]; then
+        # shellcheck disable=SC2102
+        pip install -e .[dev-pinned,pinned]
```
ztefanie, line 57: Why is this if-else needed? Our `install.sh`-script does update dependencies, when they are updated in the pyproject.toml.

ztefanie (approved):

tools/update_dependencies.sh
timobrembeck: Just a small optimization to skip the npm packages from being installed again in the `install.sh` if not needed

PR #28, opened 2023-04-19
Add djlint

JoeyStk:
### Short description
Add djlint

### Proposed changes

- Add djlint to pyproject.toml
- Add djlint to pre-commit
- Add djlint as tool
- Add djlint Circle CI job

### Side effects

- Rebasing after PR #26  has been merged

### Resolved issues

Fixes: #25

PR #32, opened 2023-04-19
Add isort

### LINKED ISSUE #29: Add isort
### Motivation

We want to avoid messy, unsorted imports to make it easier for developers to keep an overview of all importants

### User Story

As a programmer I want sorted imports so that I can keep an overview of all imports easily

### Proposed Solution

Add isort

### Alternatives
Do it manually

### Additional Context

None

### Screenshots

None

JoeyStk:
### Short description
This PR adds isort as sorter for imports

### Proposed changes

- Add isort to pyproject.toml
- Add isort as pre-commit
- Add isort as tool 
- Add isort as Circle CI job

### Side effects

- Rebasing after #28 and #26 have been merged

### Resolved issues

Fixes: #29

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,7 @@ dev = [
     "black",
     "build",
     "djlint",
+    "isort",
     "pylint",
     "pylint-django",
     "pylint-per-file-ignores",
@@ -122,6 +123,10 @@ skip-magic-trailing-comma = true
 
 [tool.djlint]
 profile="django"
+
+[tool.isort]
+profile = "black"
+order_by_type = false
 
 [tool.pylint.main]
 jobs = 0
--- a/tools/isort.sh
+++ b/tools/isort.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# This script can be used to sort the python import statements with isort.
+
+# Import utility functions
+# shellcheck source=./tools/utils/_functions.sh
+source "$(dirname "${BASH_SOURCE[0]}")/utils/_functions.sh"
+
+require_installed
+
+# Run isort
+echo "Sorting import statements with isort..." | print_info
+isort .
+echo "✔ Sorting finished" | print_success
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -105,6 +106,10 @@ ignore-paths = [
 fail-on = "I"
 django-settings-module="integreat_compass.core.settings"
 
+[tool.isort]
+profile = "black"
+order_by_type = false
```
timobrembeck: Here as well, please don't put it in between two pylint config blocks

tools/isort.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# Import utility functions
+# shellcheck source=./tools/utils/_functions.sh
+source "$(dirname "${BASH_SOURCE[0]}")/utils/_functions.sh"
+
+require_installed
+
+# Run isort
+echo "Sorting import statements with isort..." | print_info
+isort "${BASE_DIR}"
```
timobrembeck:
```suggestion
isort .
```

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dynamic = ["version"]
 [project.optional-dependencies]
 dev = [
     "build",
+    "isort",
```
timobrembeck, line 40: The pinned deps are missing

timobrembeck (changes requested):

pyproject.toml
JoeyStk: It was already added during the pylint PR and already exists

timobrembeck: ah I see, sorry!

pyproject.toml
JoeyStk: ~When sorting them alphabetically, is it okay to put them before [tool.setuptools]? Or should they be behind the setuptools and then sorted alphabetically?~

Edit: You already answered this in [this PR](https://github.com/digitalfabrik/integreat-compass/pull/26)

timobrembeck: I think both approaches are fine, although I would somehow argue that `[tool.setuptools]` is special since it is the core setting how the package is installed with pip...

timobrembeck (approved):

PR #33, opened 2023-04-20
Add Shellcheck

### LINKED ISSUE #31: Add shellcheck
### Motivation

We want to make sure that our bash scripts are valid and follow best practices for bash scripting...

### Proposed Solution

- Add shellcheck pre-commit hook
- Add shellcheck CircleCI job

ztefanie:
### Short description

Added Shellcheck to ci and pre commit hook.

### Proposed changes

Added some comments to the script to avoid shellcheck from failing

### Side effects

none

### Resolved issues

Fixes: #31

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  shellcheck: circleci/shellcheck@3.1.2
+
 jobs:
   pip-install:
     docker:
@@ -126,6 +129,12 @@ workflows:
       - pylint:
           requires:
             - pip-install
+      - shellcheck/check:
+          dir: ./tools
+          external_sources: true
+          filters:
+            branches:
+              ignore: main
   main:
     jobs:
       - pip-install:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,9 @@
 repos:
+  - repo: https://github.com/shellcheck-py/shellcheck-py
+    rev: v0.8.0.4
+    hooks:
+      - id: shellcheck
+        args: [--external-sources]
   - repo: https://github.com/pre-commit/mirrors-eslint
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -126,6 +129,12 @@ workflows:
       - pylint:
           requires:
             - pip-install
+      - shellcheck/check:
+          dir: ./tools
+          external_sources: true
```
timobrembeck, line 134:
```suggestion
```
This is probably not needed, since the utils are also checked themselves (and thus not considered an external source)?

.pre-commit-config.yaml
```diff
@@ -1,4 +1,9 @@
 repos:
+  - repo: https://github.com/shellcheck-py/shellcheck-py
+    rev: v0.8.0.4
+    hooks:
+      - id: shellcheck
+        args: [--external-sources]
```
timobrembeck, line 6: Here on the other hand I see that it might be useful when a change in the one tool (but unchanged util) causes the pre-commit hook to only get the changed file as input and the util is considered an external source...

timobrembeck (approved):
Thanks!

.circleci/config.yml
ztefanie: Shellcheck fails when I remove this, as `SC1091 (info): Not following: ./tools/utils/_functions.sh was not specified as input (see shellcheck -x).`

See: https://app.circleci.com/pipelines/github/digitalfabrik/integreat-compass/77/workflows/35ab449d-2711-43f3-a7c3-c3cfd5276f3c/jobs/250

Will keep it as it is, if you are fine with this.

timobrembeck: Oh haha, stupid me. Sorry about that, I overlooked that only passing the file path does not automatically pass all files in that directory as command line arguments as well :see_no_evil:

PR #34, opened 2023-04-20
Enable pylint for more files

timobrembeck:
### Short description

Enable pylint for more files.
Unfortunately, pylint is not able to disable specific warnings for entire directories (see https://github.com/pylint-dev/pylint/issues/618), so I added the plugin https://github.com/christopherpickering/pylint-per-file-ignores

### Proposed changes

- Enable pylint for tests directory (but disable `unused-argument` and `missing-function-docstring`)
- Enable pylint for `integreat-compass-cli`

### DIFF
--- a/integreat_compass/integreat-compass-cli
+++ b/integreat_compass/integreat-compass-cli
@@ -2,7 +2,6 @@
 """Django's command-line utility for administrative tasks."""
 import os
 import sys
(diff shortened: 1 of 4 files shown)


charludo (approved):
Thanks! Seems to work without problems.

Issue #40, opened 2023-04-20
View offer details


svenseeberg:
### Motivation
Core functionality of the tool.

### User Story
As a job center counselor I want to view the details of an offer so that I can provide necessary information (requirements, contact persons) about an offer to my client.

### Proposed Solution
- We do not have to implement all the details on the screenshot in the first phase. "IFK-Kommentare", "Erfolgsgeschichten" and "Bewertung" can be omitted. Also the menu link (3 vertical dots) is not required yet (we have no defined functionality for it).

### Alternatives
- None

### Additional Context
- None

### Screenshots
![image](https://user-images.githubusercontent.com/15197000/233360349-43e24c26-0aa0-4804-b6a8-34947e04f359.png)

charludo:
> * We do not have to implement all the details on the screenshot in the first phase. "IFK-Kommentare", "Erfolgsgeschichten" and "Bewertung" can be omitted.

~~Ahh, this was were I'd seen mention of not implementing comments for now. @svenseeberg do you remember where you got that info from?~~

Ah I see, you just extracted this to #41. Makes sense, nevermind 😄 

> Also the menu link (3 vertical dots) is not required yet (we have no defined functionality for it).

In Figma this is where the "report" button goes, should we just put it directly onto the offer details card?

PR #47, opened 2023-04-29
Add required models

### LINKED ISSUE #45: Add required models
### Motivation

Groundwork required for user story issues

### User Story

-

### Proposed Solution

Implement models corresponding to the ERD (see below)

### Screenshots

grafik

charludo:
### Short description

Add required models

### Proposed changes

- add required models
- add constants where necessary
- define some roles (these need further discussion)

### Side effects

- some of the property methods probably don't behave as they should, I had to do some guesswork 

### Resolved issues

Fixes: #45


docs/src/ref/integreat_compass.cms.constants.rst
timobrembeck, line 1: The disadvantage of having these files in the repo is that every renaming/adding/deletion of a python file also needs an update of the documentation, even if the documentation files are just autogenerated...

integreat_compass/cms/constants/allowed_media.py
```diff
@@ -0,0 +1,50 @@
...
+LEGACY_TYPES = [
+    (SVG, _("SVG image")),
+    (GIF, _("GIF image")),
+    (DOC, _("DOC document")),
+    (DOCX, _("DOCX document")),
+    (XLS, _("XLS document")),
+    (XLSX, _("XLSX document")),
+    (PPT, _("PPT document")),
+    (XLSX, _("PPTX document")),
+]
```
timobrembeck:
```suggestion
```
Since we're not migration from anywhere, I don't think we need to support these file types.

integreat_compass/cms/constants/offer_group_types.py
```diff
@@ -0,0 +1,11 @@
+"""
+This module contains all string representations of all group types in which offers
+can be offered, used by :class:`~integreat_compass.cms.models.offers.offer.Offer`
+"""
+from django.utils.translation import gettext_lazy as _
+
+PRIVATE = 0
+GROUP = 1
+BOTH = 2
```
timobrembeck:
```suggestion
PRIVATE = "PRIVATE"
GROUP = "GROUP"
BOTH = "BOTH"
```
Since the underlying model fields are `CharField`s, you have to use strings as choices for them. You could also convert them to integer fields to save a few bytes of db storage, but the advantage of having readable string constants is that it's easier to debug and you don't have to remember the meaning of the numbers when inspecting form values etc...

integreat_compass/cms/constants/offer_version_states.py
```diff
@@ -0,0 +1,15 @@
+"""
+This module contains all string representations of all approval states of offer
+versions, used by :class:`~integreat_compass.cms.models.offers.offer_version.OfferVersion`
+"""
+from django.utils.translation import gettext_lazy as _
+
+REJECTED = 0
+APPROVED = 1
+PENDING = 2
```
timobrembeck:
```suggestion
REJECTED = False
APPROVED = True
PENDING = None
```
In this case, I think boolean values make more sense for these constants...

integreat_compass/cms/constants/roles.py
```diff
@@ -0,0 +1,28 @@
...
+    (BOARD_MEMBER, _("Board Member")),
+    (INTEGRATION_SPECIALIST, _("Integration Specialist")),
+    (OFFER_PROVIDER, _("Offer Provider")),
+]
+
+BOARD_MEMBER_PERMISSIONS = ["view_offer_votes", "vote_on_offer"]
+
+INTEGRATION_SPECIALIST_PERMISSIONS = ["favorite_offer", "comment_on_offer"]
+
+OFFER_PROVIDER_PERMISSIONS = ["change_offer", "upload_mediafile"]
```
timobrembeck: The integreat-cms has proven that managing permissions in migrations is quite cumbersome, I wouldn't recommend it for a new system.
A different approach would be to manage them via fixtures, see https://github.com/digitalfabrik/integreat-cms/issues/1857

integreat_compass/cms/models/interactions/comment.py
```diff
@@ -0,0 +1,57 @@
...
+    date = models.DateTimeField(auto_now_add=True)
+    comment = models.CharField(
+        max_length=255,
+        blank=False,
+        null=False,
+        verbose_name=_("comment"),
+        help_text=_("Additional context for the given rating"),
+    )
+    rating = models.PositiveSmallIntegerField(
+        choices=((i, i) for i in range(1, 6)),
```
timobrembeck: If the rating has to be between 1 and 5, I think validators make more sense than choices.
```suggestion
        validators=[
            MinValueValidator(1),
            MaxValueValidator(5),
        ]
```
(requires `from django.core.validators import MaxValueValidator, MinValueValidator`)

integreat_compass/cms/models/offers/document.py
```diff
@@ -0,0 +1,88 @@
...
+class Document(AbstractBaseModel):
+    """
+    Data model representing a Language.
+    """
+
+    offer_version = models.ForeignKey(
+        OfferVersion, on_delete=models.CASCADE, null=False
+    )
+    name = models.CharField(max_length=255, null=False)
+    blob = models.FileField(
```
timobrembeck:
```suggestion
    file = models.FileField(
```
A bit nit picky, but `blob` usually refers to binary data saved in the database, not on the file system, right? :sweat_smile: 
In this case, the db field only contains a string with the path where the actual file is stored on the disk.

integreat_compass/cms/models/offers/offer.py
```diff
@@ -0,0 +1,92 @@
...
+    mode_type = models.CharField(
+        choices=offer_mode_types.CHOICES,
+        null=False,
+        default=offer_mode_types.IN_PERSON,
+        verbose_name=_("lesson mode"),
+        help_text=_("Select in what mode lessons are offered"),
+    )
+
+    @cached_property
+    def public_version(self):
```
timobrembeck, line 36:
```suggestion
    def approved_version(self):
```

integreat_compass/cms/models/offers/offer_version.py
```diff
@@ -0,0 +1,114 @@
...
+from ..abstract_base_model import AbstractBaseModel
+from .language import Language
+
+
+class OfferVersion(AbstractBaseModel):
+    """
+    Data model representing a Language.
+    """
+
+    offer = models.ForeignKey("Offer", on_delete=models.CASCADE)
```
```suggestion
    offer = models.ForeignKey("Offer", on_delete=models.CASCADE, related_name="versions")
```
You could set a related name here to be able to address `offer.versions` directly instead of `offer.offerversions`.

integreat_compass/cms/models/interactions/comment.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Data model representing a Comment.
+    """
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    offer_version = models.ForeignKey(OfferVersion, on_delete=models.CASCADE)
+    date = models.DateTimeField(auto_now_add=True)
+    comment = models.CharField(
+        max_length=255,
+        blank=False,
```
timobrembeck:
```suggestion
```
`blank=False` is the default anyway, so I would remove it on all model fields.

integreat_compass/cms/models/interactions/comment.py
```diff
@@ -0,0 +1,57 @@
...
+        null=False,
```
timobrembeck:
```suggestion
```
`null=False` is the default anyway, so I would remove it on all model fields.

tools/prune_database.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Remove media files (because they are no longer usable without the corresponding database entries)
+rm -rfv "${PACKAGE_DIR:?}/media"
+echo "Removed media files" | print_info
+
+if [[ -n "$INTEGREAT_COMPASS_REDIS_CACHE" ]]; then
+    # Invalidate cache if enabled
+    deescalate_privileges integreat-compass-cli invalidate all --verbosity "${SCRIPT_VERBOSITY}"
+    echo "Invalidated Redis cache" | print_info
+fi
```
timobrembeck:
```suggestion
```
I think this project does not have a cache configured

integreat_compass/core/settings.py
```diff
@@ -122,3 +126,31 @@
...
+#: The maximum size of media files in bytes
+MEDIA_MAX_UPLOAD_SIZE = int(
+    os.environ.get("INTEGREAT_COMPASS_MEDIA_MAX_UPLOAD_SIZE", 3 * 1024 * 1024)
+)
+
+# Voting system settings
+
+#: Maximum votes cast on new offers
+NEW_OFFER_GREMIUM_SIZE = int(
+    os.environ.get("INTEGREAT_COMPASS_NEW_OFFER_GREMIUM_SIZE", "6")
```
timobrembeck:
```suggestion
    os.environ.get("INTEGREAT_COMPASS_NEW_OFFER_GREMIUM_SIZE", 6)
```
`int()` also works when the number already is an int.

integreat_compass/core/settings.py
```diff
@@ -122,3 +126,31 @@
...
+)
+
+#: Maximum votes cast on changed offers
+CHANGED_OFFER_GREMIUM_SIZE = int(
+    os.environ.get("INTEGREAT_COMPASS_CHANGED_OFFER_GREMIUM_SIZE", "4")
```
timobrembeck:
```suggestion
    os.environ.get("INTEGREAT_COMPASS_CHANGED_OFFER_GREMIUM_SIZE", 4)
```

integreat_compass/cms/models/users/role.py
```diff
@@ -0,0 +1,63 @@
...
+from django.db import models
+from django.utils import translation
+from django.utils.functional import cached_property
+from django.utils.translation import gettext_lazy as _
+
+from ...constants import roles
+from ..abstract_base_model import AbstractBaseModel
+
+
+class Role(AbstractBaseModel):
```
timobrembeck: I'm not even sure if this project needs the proxy model `Role`, maybe the default Django `Group` is enough?

timobrembeck (commented):
Thanks a lot! :muscle: 
Didn't have time to fully review this, but already have a few preliminary comments.

integreat_compass/cms/constants/allowed_media.py
charludo: Calling them legacy types is probably unnecessary, but shouldn't they be allowed? I can absolutely see some provider attempting to upload a pptx as their additional course info material

integreat_compass/cms/models/offers/document.py
charludo: Yes, true. I took the name from the ERD, but agree file is clearer

integreat_compass/cms/models/offers/offer.py
charludo: Hm... There can be many approved versions. And I think it makes intuitive sense to call the version that should be displayed to visitors the public version.

integreat_compass/cms/models/users/role.py
charludo: We should probably discuss roles tomorrow in the team call. I'm more than open to a simpler system, as you mentioned above. We had not really talked about it previously, I just felt the need to include at least a rough draft with this PR.

Btw, thanks for the (preliminary) review, I know it's a frustratingly large PR, sorry about that.

integreat_compass/cms/constants/allowed_media.py
timobrembeck: These document types are not ideal to distribute information, because they require additional apps to be correctly displayed. It would be better to convert these documents to PDF before uploading them.

integreat_compass/cms/models/offers/offer.py
timobrembeck: Ok, nevermind :+1:

integreat_compass/cms/models/offers/offer_version.py
charludo: Hold on, isn't this already given by:

```
class Meta:
        ordering = ["offer_version_date"]
        default_permissions = ("change", "delete", "view")
```

timobrembeck: No, the related name is the name of the attribute under which the reverse relationship is available...

https://docs.djangoproject.com/en/4.2/topics/db/queries/#backwards-related-objects
https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.ForeignKey.related_name

timobrembeck: Ah, you mean the `default_related_name`? Yes, this covers the same relationship, but `offer.versions` is still more concise than `offer.offer_versions`...
But yeah, doesn't make a huuuge difference :sweat_smile:

charludo: > Ah, you mean the `default_related_name`? Yes, this covers the same relationship, but `offer.versions` is still more concise than `offer.offer_versions`... But yeah, doesn't make a huuuge difference sweat_smile

Yes :D

We could just rename the `default_related_version` to `versions`... It would probably still be clear enough...? Currently it's only relevant when accessing versions from the offer anyways.

timobrembeck: Hmm, the problem with that is that the default related name automatically applies to all foreign keys, so also `language` in this case...
And `language.versions` could be misleading, since it implies that it's a version of the language itself...
So I would suggest that the default related name is foreign-key-agnostic in a sense that it contains the whole model name, e.g. `language.offer_versions`, and only use the short related name `offer.versions` for the offer foreign key to avoid the duplication of the term "offer"...

charludo: Alright, makes sense.

integreat_compass/cms/models/interactions/comment.py
```diff
@@ -0,0 +1,55 @@
...
+class Comment(AbstractBaseModel):
+    """
+    Data model representing a Comment.
+    """
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    offer_version = models.ForeignKey(OfferVersion, on_delete=models.CASCADE)
+    date = models.DateTimeField(auto_now_add=True)
+    comment = models.CharField(
+        max_length=255,
```
svenseeberg: When looking at the example texts in #40 I think 255 characters is too short for a comment. 512 or 1024 could be better?

svenseeberg (changes requested):
Looks good to me, just one minor note.

integreat_compass/cms/models/interactions/comment.py
timobrembeck: Maybe even a `TextField`? Then we can control the max length in the form more dynamically and don't need a db migration when the requirements change...

integreat_compass/cms/constants/offer_mode_types.py
```diff
@@ -0,0 +1,11 @@
+"""
+This module contains all string representations of all mode types in which offers
+can be offered, used by :class:`~integreat_compass.cms.models.offers.offer.Offer`
+"""
+from django.utils.translation import gettext_lazy as _
+
+ONLINE = 0
+HYBRID = 1
+IN_PERSON = 2
```
timobrembeck: Here, the same applies like in https://github.com/digitalfabrik/integreat-compass/pull/47#discussion_r1182297417... The field is a char field, so the choices should also be chars.

integreat_compass/cms/constants/organization_types.py
```diff
@@ -0,0 +1,15 @@
+"""
+This module contains all string representations of all types of organizations,
+used by :class:`~integreat_compass.cms.models.organizations.organization.Organization`
+"""
+from django.utils.translation import gettext_lazy as _
+
+NON_PROFIT = 0
+FOR_PROFIT = 1
+EV = 2
```
timobrembeck: And here as well

integreat_compass/cms/models/offers/offer.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @cached_property
+    def comments(self):
+        """
+        Method to retrieve all comments on an Offer.
+
+        :return: List of comments together with information on whether the comment was made on the latest offer version
+        :rtype: list [ dict [ ~integreat_compass.cms.models.interactions.comment.Comment, bool ] ]
+        """
+        versions = self.offerversions.order_by("-version_date")
```
```suggestion
        versions = self.versions.order_by("-version_date")
```
I think here we can now use the new related name, right?

integreat_compass/cms/constants/roles.py
timobrembeck, line 1: If we don't go via the proxy model `Role`, I think we can rename this constant to `groups.py` or `group_names.py` so we don't mix up the two terms if they mean the same thing.

integreat_compass/cms/models/users/user.py
```diff
@@ -0,0 +1,135 @@
...
+    @cached_property
+    def role(self):
+        """
+        We refer to Django user groups as roles.
+
+        :return: The role of this user
+        :rtype: ~integreat_compass.cms.models.users.role.Role
+        """
+        return self.groups.first().name if self.groups.exists() else None
+
```
timobrembeck:
```suggestion
```
If we don't use the `Role` mode, I guess we can just reference `user.groups` directly?

integreat_compass/cms/models/users/user.py
```diff
@@ -0,0 +1,135 @@
...
+    def get_repr(self):
+        """
+        This overwrites the default Django ``__repr__()`` method which would return ``<User: User object (id)>``.
+        It is used for logging.
+
+        :return: The canonical string representation of the user
+        :rtype: str
+        """
+        return f"<User (id: {self.id}, display_name: {self.display_name}, role: {self.role}, is_staff: {self.is_staff})>"
```
timobrembeck:
```suggestion
        fields = [
            f"id: {self.id}",
            f"email: {self.email}",
            f"display_name: {self.display_name}",
            f"group: {self.groups.first()}",
            f"is_staff: {self.is_staff}",
        ]
        return f"<User ({', '.join(fields)})>"
```
We could show the email as well here

integreat_compass/cms/constants/roles.py
timobrembeck: Also, we need to modify the `Group` model in order to make use of the translated name, e.g. put 
```
from django.contrib.auth.models import Group
from django.utils.translation import gettext as _

Group.__str__ = lambda self : _(self.name)
```
somewhere where it's loaded after start, e.g. in [AppConfig.ready()](https://docs.djangoproject.com/en/4.2/ref/applications/#django.apps.AppConfig.ready)

timobrembeck (commented):
Another few nit-picky comments...

But great work, really, this is probably much closer to the final implementation than most initial PRs of other projects :grin:

integreat_compass/cms/models/users/user.py
charludo: Hm... It might still be handy to have this method, maybe we can just rename it? Or is there a way around having to do `self.groups.first().name if self.groups.exists() else None` to get the role/group otherwise?

timobrembeck: Well, we could also just rely on the fact that `str(group)` is converted to `group.name` and just use `self.groups.first()` directly...

After all, `self.groups.first().name if self.groups.exists() else None` also makes two sql queries if the group exists, which is not really ideal... 

I can imagine that even this small property saves a little bit of typing:

```suggestion
    @cached_property
    def group(self):
        """
        Return the primary group of this user

        :return: The first group of this user
        :rtype: ~django.contrib.auth.models.Group
        """
        return self.groups.first()

```

timobrembeck: Well, we could also just rely on the fact that `str(group)` is converted to `group.name` and just use `self.groups.first()` directly...

After all, `self.groups.first().name if self.groups.exists() else None` also makes two sql queries if the group exists, which is not really ideal... 

I can imagine that even this small property saves a little bit of typing:

```suggestion
    @cached_property
    def group(self):
        """
        Return the primary group of this user

        :return: The first group of this user
        :rtype: ~django.contrib.auth.models.Group
        """
        return self.groups.first()

```

charludo: Sounds good to me!

timobrembeck (approved):
:tada:

integreat_compass/cms/apps.py
```diff
@@ -0,0 +1,22 @@
...
+
+    :param name: The name of the app
+    :type name: str
+    """
+
+    name = "integreat_compass.cms"
+    verbose_name = _("CMS")
+
+    def ready(self):
+        # pylint: diable-next=import-outside-toplevel
```
timobrembeck:
```suggestion
        # pylint: disable=import-outside-toplevel
```

svenseeberg (approved):
Thanks!

PR #48, opened 2023-05-03
Add versioning tool

### LINKED ISSUE #10: Add versioning tool
- Decide on a versioning scheme, e.g. CalVer `YYYY.MM.INC0[-TAG]`
- Add versioning tool that is able to automatically bump the version, e.g. https://github.com/mbarkhau/bumpver

JoeyStk:
### Short description
This PR introduced CalVer as versioning scheme, and adds bumpver as versioning tools, that automatically bumps the version

### Proposed changes

- Add bumpver
- Define CalVer as versioning scheme in setup for bumpver
- Set current version to 2023.0.0

### Side effects

I read about versioning schemes for the first time today, but still was trying to make an informed decision. Here is my reasoning for why I chose to go with a calendar based versioning scheme:
I think one could make a case for logical schemes, like MAJOR.MINOR.PATCH, but I see two issues with it: 1. AFAIK it's still quite unclear how this project will evolve, and 2. I don't really see any cases where we would bump the major version. Therefore I came to the conclusion that a more flexible scheme, such as CalVer, fits our needs better. Please feel free to comment or suggest different schemes. 

### Resolved issues

Fixes: #10

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,12 +16,12 @@ repos:
       - id: no-commit-to-branch
         args: [--branch, main, --branch, develop]
   - repo: https://github.com/shellcheck-py/shellcheck-py
-    rev: v0.8.0.4
+    rev: v0.9.0.2
     hooks:
       - id: shellcheck
         args: [--external-sources]
   - repo: https://github.com/pre-commit/mirrors-eslint
-    rev: v8.27.0
+    rev: v8.39.0
     hooks:
       - id: eslint
         files: \.[jt]sx?$
@@ -40,7 +40,7 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/Riverside-Healthcare/djLint
-    rev: v1.23.1
+    rev: v1.25.0
     hooks:
       - id: djlint-django
   - repo: https://github.com/pycqa/isort
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ dynamic = ["version"]
 dev = [
     "black",
     "build",
+    "bumpver",
     "djlint",
     "isort",
     "pylint",
@@ -54,28 +55,29 @@ dev = [
 ]
 pinned = [
     "asgiref==3.6.0",
-    "Django==4.2",
-    "psycopg==3.1.8",
-    "psycopg-binary==3.1.8",
+    "Django==4.2.1",
+    "psycopg==3.1.9",
+    "psycopg-binary==3.1.9",
     "sqlparse==0.4.4",
     "typing_extensions==4.5.0",
 ]
 dev-pinned = [
     "alabaster==0.7.13",
-    "astroid==2.15.3",
+    "astroid==2.15.4",
     "Babel==2.12.1",
     "black==23.3.0",
     "build==0.10.0",
+    "bumpver==2022.1120",
     "certifi==2022.12.7",
     "charset-normalizer==3.1.0",
     "click==8.1.3",
     "colorama==0.4.6",
-    "coverage==7.2.3",
+    "coverage==7.2.5",
     "cssbeautifier==1.14.7",
     "dill==0.3.6",
-    "djlint==1.23.1",
-    "EditorConfig==0.12.3",
+    "djlint==1.25.0",
     "docutils==0.18.1",
+    "EditorConfig==0.12.3",
     "exceptiongroup==1.1.1",
     "execnet==1.9.0",
     "html-tag-names==0.1.2",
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -150,6 +150,18 @@ version = { attr = "integreat_compass.__version__" }
 [tool.setuptools.packages.find]
 include = ["integreat_compass*"]
 
+[tool.bumpver]
+current_version = "2023.0.0"
```
timobrembeck:
```suggestion
current_version = "2023.3.0"
```
0 is not a valid month. Please keep the current version and only bump it when there is a real release. Merging this PR does not create a new release, so it should also not change the version.

pyproject.toml
```diff
@@ -150,6 +150,18 @@ version = { attr = "integreat_compass.__version__" }
...
+version_pattern = "YYYY.MM.INC0[-TAG]"
+commit_message = """
+Bump version to {new_version}
+
+[tool.bumpver.file_patterns]
+"pyproject.toml" = [
+    'current_version = "{version}"',
+    'version = "{version}"',
```
timobrembeck:
```suggestion
```

This pattern does not exist in `pyproject.toml`. Instead, the version is dynamically defined here (so the `__init__.py` needs to be listed in `file_patterns`):
https://github.com/digitalfabrik/integreat-compass/blob/75b3e26d507f828d6cde4206a76fdc10f8b58633/integreat_compass/__init__.py#L6

pyproject.toml
```diff
@@ -150,6 +150,18 @@ version = { attr = "integreat_compass.__version__" }
 [tool.setuptools.packages.find]
 include = ["integreat_compass*"]
 
+[tool.bumpver]
```
timobrembeck, line 156: Did you forget to add `bumpver` as dev dependency?

timobrembeck (changes requested):

pyproject.toml
```diff
@@ -150,6 +150,18 @@ version = { attr = "integreat_compass.__version__" }
...
+current_version = "2023.0.0"
+version_pattern = "YYYY.MM.INC0[-TAG]"
+commit_message = """
+Bump version to {new_version}
```
timobrembeck, line 160:
```suggestion
Bump version to {new_version}
"""

```
And the multi-line string needs to have an ending as well

pyproject.toml
JoeyStk: Oh thank you, that explains a lot!

pyproject.toml
JoeyStk: Yes, I had. It's added now

pyproject.toml
```diff
@@ -55,27 +56,28 @@ dev = [
...
     "charset-normalizer==3.1.0",
     "click==8.1.3",
     "colorama==0.4.6",
-    "coverage==7.2.3",
+    "coverage==7.2.5",
     "cssbeautifier==1.14.7",
     "dill==0.3.6",
-    "djlint==1.23.1",
-    "EditorConfig==0.12.3",
+    "djlint==1.25.0",
```
timobrembeck, line 78: When you update djlint, could you also update the corresponding pre-commit hook?

timobrembeck (changes requested):

pyproject.toml
timobrembeck: This is not yet resolved. The version in `__init__.py` is still missing from the file patterns.

.pre-commit-config.yaml
```diff
@@ -32,15 +32,15 @@ repos:
           - "@typescript-eslint/eslint-plugin@5.42.1"
           - "eslint-plugin-prefer-arrow@1.2.3"
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v2.7.1
+    rev: v3.0.0-alpha.9-for-vscode
```
timobrembeck:
```suggestion
    rev: v2.7.1
```
Please do not blindly execute scripts, but also read their output:

> Updating pre-commit hooks...
> Note: Please downgrade any alpha versions manually

The problem is that `pre-commit autoupdate` always takes the latest version, even if they're alpha or beta versions... I did not find a way around it unfortunately...

pyproject.toml
```diff
@@ -150,6 +153,22 @@ version = { attr = "integreat_compass.__version__" }
...
+
+
+[tool.bumpver.file_patterns]
+"pyproject.toml" = [
+    'current_version = "{version}"',
+]
+"integreat_compass/__init__.py" = [
+    '__version__ = "{version}"',
+]
+
```
timobrembeck, line 176:
```suggestion
"docs/src/conf.py" = [
    'release = "{version}"',
]

```
Could you also fix the version in this file from `2023.4.0` to `2023.3.0`?

timobrembeck (changes requested):

.pre-commit-config.yaml
JoeyStk: I'm sorry!

pyproject.toml
JoeyStk: This should be applied now

pyproject.toml
```diff
@@ -150,6 +153,25 @@ version = { attr = "integreat_compass.__version__" }
 [tool.setuptools.packages.find]
 include = ["integreat_compass*"]
 
+[tool.bumpver]
+current_version = "2023.3.0"
+version_pattern = "YYYY.MM.INC0[-TAG]"
+commit_message = """
+Bump version to {new_version}
+"""
```
timobrembeck, line 163:
```suggestion

[skip ci]
"""
```
It' important that version bumps do not trigger the CI workflow, otherwise we run into an indefinite loop where a merge on main triggers a version bump, and the version bump triggers another version bump, and so on...

timobrembeck (changes requested):

pyproject.toml
JoeyStk: Ah, thank you for the explanation. I didn't realize this was part of the issue

timobrembeck: Maybe not necessarily part of this issue, but easy to miss later on

timobrembeck (approved):
:rocket:

JoeyStk:
Thank you very much for all your help. I know it must be (at least in parts) tiresome

PR #49, opened 2023-05-08
Add base templates, tailwind config and landing page

charludo:
### Short description

I noticed these were not setup yet, and we did not have an issue for it. I hope this will save a fair amount of time once we can start realizing the user story issues.

### Proposed changes

- add base templates
- add tailwind config
- add index view
- add url pattern files

### Side effects

- 
- 

### Resolved issues

Fixes: n/a


integreat_compass/cms/templates/_base.html
```diff
@@ -0,0 +1,33 @@
...
+                     alt="{% translate "Logo" %}"
+                     class="h-24 w-48">
+            </a>
+            <a href="#" class="ml-6 mr-4">{% translate "Dashboard" %}</a>
+            <a href="#" class="ml-6 mr-4">{% translate "All offers" %}</a>
+            {% if perms.cms.edit_offer %}
+                <a href="#" class="ml-6 mr-4">{% translate "My Offers" %}</a>
+                <a href="#" class="ml-6 mr-4">{% translate "Add new offer" %}</a>
+            {% endif %}
+            <div class="ml-auto mr-4 flex-align-right">
```
JoeyStk:
```suggestion
            <div class="ml-auto mr-4 flex-align-right font-bold">
```

integreat_compass/cms/templates/_base.html
```diff
@@ -0,0 +1,33 @@
+{% extends "_raw.html" %}
+{% load i18n %}
+{% load static %}
+{% block raw_content %}
+    <div class="flex justify-between items-center shadow-md bg-white z-50 fixed w-screen top-0">
+        <div class="container mx-auto flex items-center font-medium text-xl text-gray-700 hover:text-gray-900">
```
JoeyStk: I see two issues here:
1. I think text-gray-700 is too dark. It translates to #374151 but the design suggests to use #757575. I see two options how to go about this: 
- Overwrite text-gray-700 in the config to #757575
- Use text-gray-500 instead, since it's closer to #757575

And we should already have in mind, that there will be paragraphs that are suggested to have #595959 as font color.

2. Due to the fact that the hover-effect is on the div that's wrapping all links, we have the behavior, that when you hover over one link all links get highlighted. I don't think this is intended. Instead I think the intended behavior is that only the link that gets hovered on is highlighted.

tailwind.config.js
```diff
@@ -0,0 +1,80 @@
...
+          400: "#ecf1fd",
+          500: "#e4ebfc",
+          600: "#cdd4e3",
+          700: "#abb0bd",
+          800: "#898d97",
+          900: "#70737b",
+        },
+        background: {
+          50: "#ebe5e5",
+          900: "#0a001167",
```
JoeyStk:
```suggestion
          900: "#0a0011ab",
```
I think while converting the opacity to hexadecimal something got mixed-up :)

integreat_compass/cms/templates/index.html
```diff
@@ -0,0 +1,16 @@
...
+{% block content %}
+    <div class="absolute top-0 right-0 bottom-0 left-0 flex justify-center items-center bg-background-900 z-1">
+        <div class="flex flex-col justify-center gap-4 text-center">
+            <h1 class="font-bold text-5xl drop-shadow-md mb-12 text-white">{% translate "Arriving, reimagined." %}</h1>
+            <a href="#"
+               class="bg-redMuted hover:bg-red-500 text-white font-bold text-3xl py-4 px-4 rounded-2xl">{% translate "Discover Integreat Compass" %}</a>
+            <a href="#"
+               class="bg-white hover:bg-gray-200 font-bold text-3xl py-4 px-4 rounded-2xl">{% translate "Add new offer" %}</a>
+        </div>
+    </div>
```
JoeyStk, line 17: Maybe this is out of cope for this PR, but I think we could already think about which icon-library we want to use and could already the chevron-down that is displayed in the design. If this is out of scope, I would suggest we open a new issue for this

JoeyStk (changes requested):
This already looks amazing!
I only took a closer look at the graphics, so it would be great if someone else could take a look at the other aspects of this PR. 
About the UI please see my suggestions below.
Thank you a lot for the great work!

integreat_compass/cms/templates/_base.html
charludo: > I see two issues here:
> 
>     1. I think text-gray-700 is too dark. It translates to #374151 but the design suggests to use #757575. I see two options how to go about this:
> 
> 
>     * Overwrite text-gray-700 in the config to #757575
> 
>     * Use text-gray-500 instead, since it's closer to #757575
> 

Yep, I'll switch it out for 500. Are we intending to match all the colors (esp. greys) exactly? I find some of them a little lacking in contrast tbh.

> And we should already have in mind, that there will be paragraphs that are suggested to have #595959 as font color.
> 
>     2. Due to the fact that the hover-effect is on the div that's wrapping all links, we have the behavior, that when you hover over one link all links get highlighted. I don't think this is intended. Instead I think the intended behavior is that only the link that gets hovered on is highlighted.

Yep, my mistake, thanks for noticing :)

tailwind.config.js
charludo: Oh, whoops :D

integreat_compass/cms/templates/_base.html
JoeyStk: > Are we intending to match all the colors (esp. greys) exactly?

Good question. I was thinking about it too. I will ask about it

integreat_compass/cms/templates/index.html
charludo: Good idea.

Though I actually have some questions about the start page - is it even necessary? What is the advantage of using it instead of just showing the dashboard directly?

I also do not get the chevron down, is the start page supposed to be reachable by scrolling up from the dashboard?

No matter how I think about it, this just feels like an additional click and like it is not offering anything of value.

JoeyStk: Yes, I agree. That was the second point on my list to ask later. It also doesn't make a lot of sense to me, but I was giving the designer the benefit of the doubt and thought that maybe we're not seeing something :)

charludo: > we could already think about which icon-library we want to use

TBH I would just go with lucide again, those have been working pretty flawlessly afaik?

JoeyStk: Yes, I don't know of any issues either and would also just go with lucide again

charludo: @JoeyStk I know it's not required anymore, but I've left the index page in as a placeholder until the Dashboard and/or all offers views are created.

Icons are now working as well.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dev = [
     "black",
     "build",
     "bumpver",
+    "django-webpack-loader",
```
timobrembeck: I think the webpack loader is in fact a prod dependency, since it's required to parse the webpack stats and derive the file names of the bundled files from it...

integreat_compass/cms/views/index.py
```diff
@@ -0,0 +1,14 @@
...
+    """
+    Start page of the project
+
+    :param request: The current request
+    :type request: ~django.http.HttpRequest
+
+    :return: The rendered template response
+    :rtype: ~django.template.response.TemplateResponse
+    """
+    return render(request, "index.html")
```
timobrembeck:
```suggestion
from django.views.generic import TemplateView


class IndexView(TemplateView):
    """
    Start page of the project
    """

    template_name = "index.html"
```
I know it's just a placeholder, but I think it would be cool to directly use Django's class-based views here to encourage the usage of class-based views in general when more views are added :sweat_smile:

integreat_compass/cms/urls/public.py
```diff
@@ -0,0 +1,5 @@
+from django.urls import path
+
+from ..views import index
+
+urlpatterns = [path("", index.get, name="index")]
```
timobrembeck:
```suggestion
urlpatterns = [path("", index.IndexView.as_view(), name="index")]
```

integreat_compass/cms/urls/protected.py
```diff
@@ -0,0 +1,4 @@
+"""
+URLconf for login-protected views of the cms package.
+"""
+urlpatterns = []
```
timobrembeck, line 5:
```suggestion

# TODO: Add middleware to actually enforce access control on these views!

urlpatterns = []
```
I'd point out that these views are not proteced (yet)...

timobrembeck (changes requested):
Very cool! Just a few minor comments...

integreat_compass/cms/urls/protected.py
charludo: Haha, fair point... 😅

integreat_compass/cms/urls/protected.py
```diff
@@ -0,0 +1,7 @@
+"""
+URLconf for login-protected views of the cms package.
+"""
+
+# TODO: Add middleware to actually enforce access control on these views!
```
timobrembeck: Oh haha I wasn't aware that pylint fails with a TODO warning...
I guess we could also add the access control directly:
https://github.com/digitalfabrik/integreat-cms/blob/c0bbd6c40708aac4695d458906876286a6ced4d5/integreat_cms/core/middleware/access_control_middleware.py

charludo: ...oh, my solution was to just remove the word TODO, haha.

charludo: But sure, we might as well.

timobrembeck (approved):
:rocket:

integreat_compass/cms/urls/protected.py
charludo: I have added a somehat trimmed down version of the one you linked. Hope it's OK like this. Also wondering if we should add a public app or find a different way to differentiate between which routes require auth and which do not

timobrembeck: I guess this depends on the general layout of the app - if the public pages are independent from the password protected pages, then yes it makes sense to put them in their own app, but if they are in a way part of the same functionality, then I'd argue that they should live in the same app...
I'm not really familiar with the thematic parts of the project, so do what you feel makes sense.

integreat_compass/core/middleware/access_control_middleware.py
```diff
@@ -0,0 +1,43 @@
...
+logger = logging.getLogger(__name__)
+
+
+# pylint: disable=too-few-public-methods
+class AccessControlMiddleware:
+    """
+    Middleware class that performs a basic access control. For urls that are whitelisted (see
+    :attr:`~integreat_compass.core.middleware.access_control_middleware.AccessControlMiddleware.whitelist`), no additional
+    rules are enforced.
+    For all other urls, the user has to be either superuser or needs access to the current region.
```
timobrembeck:
```suggestion
    For all other urls, the user has to be logged in.
```

integreat_compass/cms/urls/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+For more information on this file, see :doc:`django:topics/http/urls`.
+"""
+
+from django.urls import include, path
+
+#: The url patterns of this module (see :doc:`django:topics/http/urls`)
+urlpatterns = [
+    path("", include("integreat_compass.cms.urls.public")),
+    path("", include("integreat_compass.cms.urls.protected")),
+]
```
timobrembeck:
```suggestion


#: The namespace for this URL config (see :attr:`django.urls.ResolverMatch.app_name`)
app_name = "cms"

#: The url patterns of this module (see :doc:`django:topics/http/urls`)
urlpatterns = [
    path("", include("integreat_compass.cms.urls.public", namespace="public")),
    path("", include("integreat_compass.cms.urls.protected", namespace="protected")),
]
```
I guess we could also make use of namespaces here and make the distinction based on it? But didn't try this out yet...
https://docs.djangoproject.com/en/4.2/topics/http/urls/#url-namespaces

charludo: Sounds like a good idea to me!

timobrembeck: Small caveat seems to be that urls have to be referenced explicitly now, so `{% url "index" %}` no longer works, instead we have to write `{% url "cms:public:index" %}`. But at the same time this makes it more robust against mistakes and the access control looks far cleaner now, so I think it's worth it...

timobrembeck (approved):
:tada:

integreat_compass/cms/urls/__init__.py
charludo: Ah, true. But yes, not really a dealbreaker.

JoeyStk (approved):
I think this is good for now. We'll probably have to do some adjustments to the UI once the re-design process is finished. But I think until then it's more important that this PR doesn't block any other features and I will open a new issue once we have the feedback from UI/UX

PR #51, opened 2023-05-20
Add Offer Application forms and view

### LINKED ISSUE #37: Offer Application
### Motivation
Offers are not immediately available to consumers (job center counselors). They first need to be admitted by a board (gate keeper). Organizations therefore need to apply with an offer.

### User Story
As a organization representative I want enter new offer ("Maßnahme") so that the offer can be found by job center employees and recommended to job center clients.

### Proposed Solution
- Add a form (login required) to enter details of an offer (title, description, contact, categories).
- The contact information has to be entered for each measure (offer) individually. The organisation information can be automatically completed from a list of existing organizations (no personal data).
- The success message should contain a message that the application "will be reviewed by the board of trustees".
- The location is excluded in this task (extra task #50 ).
- "Gesellschaftsform" should be replaced with "free of charge".

### Screenshots
image

charludo:
### Short description

Add Offer Application forms and view

### Proposed changes

- added Offer Application forms and view
- modified models (over-wrote initial migration since these are basically fixes of stuff I should have realized when implementing them initially, e.g. unique field names)
- added login, logout functionality (very basic, needs to be improved upon)
- added `test_data.json` fixture content (root user, tags, languages, example offers)
- input element and button styles

### Side effects

- hopefully not too many... 

### Resolved issues

Fixes: #37


integreat_compass/cms/fixtures/test_data.json
```diff
@@ -1 +1,1087 @@
...
+      "cost": 250,
+      "language": 2
+    }
+  },
+  {
+    "model": "cms.user",
+    "pk": 1,
+    "fields": {
+      "password": "pbkdf2_sha256$600000$PA2SmRUyUUxufQl205scvB$BlFZuqJHPxUQwRnEGsuExRR5dYUoPXD7V559Jcfczrw=",
```
ztefanie: What would this be decoded?

charludo: Ah whoops, sorry. It's `root1234`

integreat_compass/cms/forms/custom_model_form.py
```diff
@@ -0,0 +1,188 @@
...
+        try:
+            super().__init__(**kwargs)
+        except ValueError as e:
+            raise TypeError("CustomModelForm cannot be instantiated directly.") from e
+
+        # Set additional instance attributes
+        for name, value in additional_instance_attributes.items():
+            setattr(self.instance, name, value)
+
+        # Dynamically initialize logger to get module name of actual form
```
ztefanie: I know this file is mostly copied from integreat, but I think in consideration of clean code principles, this function should
1) be way shorter, therefore please split it up
2) only accomplish ONE task, so e.g. move the logging stuff to a separate funtion

And in general all these comments are really bad practice. I think we should remove them and have code that is self-explanatory.

integreat_compass/cms/forms/custom_model_form.py
```diff
@@ -0,0 +1,188 @@
...
+            *[getattr(self, attribute) for attribute in attributes],
+        )
+
+        for field_name, field in self.fields.items():
+            # If form is disabled, disable all form fields
+            if disabled:
+                field.disabled = True
+
+            # Set placeholder for every text input fields
+            if isinstance(
```
ztefanie, line 61: Setting placeholder texts can easily be moved to a own function with a proper naming and this comment can be deleted. The init function will then be much more readable.

integreat_compass/cms/forms/custom_model_form.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def clean(self):
+        """
+        This method extends the default ``clean()``-method of the base :class:`~django.forms.ModelForm` to provide debug
+        logging
+
+        :return: The cleaned data (see :ref:`overriding-modelform-clean-method`)
+        :rtype: dict
+        """
+        # Validate ModelForm
```
ztefanie: This comment has really no information and should be deleted.

integreat_compass/cms/forms/custom_model_form.py
```diff
@@ -0,0 +1,188 @@
...
+        cleaned_data = super().clean()
+        self.logger.debug(
+            "%s validated with cleaned data %r", type(self).__name__, cleaned_data
+        )
+        return cleaned_data
+
+    def save(self, commit=True):
+        """
+        This method extends the default ``save()``-method of the base :class:`~django.forms.ModelForm` to provide debug
```
ztefanie: Here we see why too many comments are harming, as this comment is just wrong. The comment suggests that this function only adds logging, but this function also adds the creator. Please update this comment

integreat_compass/cms/forms/custom_model_form.py
```diff
@@ -0,0 +1,188 @@
...
+        :return: The saved object returned by :ref:`django:topics-modelform-save`
+        :rtype: object
+        """
+        self.logger.debug(
+            "%s saved with changed data %r", type(self).__name__, self.changed_data
+        )
+        # Update submitted data with cleaned data if changes are saved
+        self.data = self.cleaned_data
+
+        # Save ModelForm, add creator if required
```
ztefanie: this comment also does not correctly describe whats going on here.

integreat_compass/cms/models/interactions/comment.py
```diff
@@ -12,7 +12,7 @@ class Comment(AbstractBaseModel):
     Data model representing a Comment.
     """
 
-    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    creator = models.ForeignKey(User, on_delete=models.CASCADE)
```
ztefanie: Does this really make what we want? If a user is deleted, I think his/her comments should stay. :thinking:

integreat_compass/cms/models/interactions/vote.py
```diff
@@ -11,7 +11,7 @@ class Vote(AbstractBaseModel):
     Data model representing a Vote.
     """
 
-    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    creator = models.ForeignKey(User, on_delete=models.CASCADE)
```
ztefanie: I think if a user is deleted, his/her votes should definitely stay, as otherwise if someone is removed from the "gremium" some offers do not have enough votes anymore.

integreat_compass/cms/models/offers/tag.py
```diff
@@ -9,7 +9,7 @@ class Tag(AbstractBaseModel):
     Data model representing a Tag.
     """
 
-    title = models.CharField(
+    tag = models.CharField(
```
ztefanie: Why have you decided to rename this? I think title is a good naming, as if we want to add something more later, e.g. category, description or whatever, title would be a appropriate naming. Also now i think "title" is better than "tag".

integreat_compass/cms/models/organizations/contact.py
```diff
@@ -11,16 +11,16 @@ class Contact(AbstractBaseModel):
     """
 
     creator = models.ForeignKey(User, on_delete=models.CASCADE, null=False)
-    name = models.CharField(
+    contact_name = models.CharField(
```
ztefanie: Why these renamings? It is unnecessary to prefix all attributes with the name of the class. Please undo.

integreat_compass/cms/models/organizations/organization.py
```diff
@@ -12,7 +12,7 @@ class Organization(AbstractBaseModel):
     """
 
     creator = models.ForeignKey(User, on_delete=models.CASCADE, null=False)
-    name = models.CharField(
+    org_name = models.CharField(
```
ztefanie: Same here, prefixing attributes with the class name seems bad.

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        offer_instance = Offer.objects.filter(id=kwargs.get("offer_id")).first()
+        if not request.user.has_perm("cms.edit_offer", offer_instance):
+            raise PermissionDenied(
+                f"{request.user!r} does not have the permission to edit {offer_instance!r}"
+            )
+
+        user = User.objects.get(pk=request.user.id)
```
ztefanie: Why do we need to load the user? Why is request.user not enough?

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+                "offer_version_form": offer_version_form,
+                "contact_form": contact_form,
+                "location_form": location_form,
+                "organization_form": organization_form,
+            },
+        )
+
+    # pylint: disable=too-many-locals
+    @transaction.atomic
+    def post(self, request, *args, **kwargs):
```
ztefanie, line 86: This function is to big, please split it up.

There are different task this functions fullfills:
1) Check permissions
2) Load User and append to different forms
3) Validate forms
4) Save forms

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+            offer_version_form,
+            contact_form,
+            location_form,
+            organization_form,
+        )
+        if not all(form.is_valid() for form in forms):
+            for form in forms:
+                form.add_error_messages(request)
+        else:
+            print(contact_form.is_valid())
```
ztefanie: Why do you use print here instead of the logger?

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+            contact_instance = contact_form.save()
+            location_instance = location_form.save()
+            organization_instance = organization_form.save()
+
+            offer_instance = offer_form.save(commit=False)
```
ztefanie: I think this is all done, so the foreign key constraints are not violated? Isn't there a better way to achieve this? Some django-function?

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+                **self.get_context_data(**kwargs),
+                "offer_form": offer_form,
+                "offer_version_form": offer_version_form,
+                "contact_form": contact_form,
+                "location_form": location_form,
+                "organization_form": organization_form,
+            },
+        )
+
+    # pylint: disable=too-many-locals
```
ztefanie: pylint error should be fixed and this should be removed.

ztefanie (changes requested):

integreat_compass/cms/fixtures/test_data.json
ztefanie: Maybe we should document this somewhere?

timobrembeck: I guess you already documented this in https://github.com/digitalfabrik/integreat-compass/pull/18/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R33?
But yes, we could also add it to the [documentation](https://integreat-compass.readthedocs.io/en/latest/dev-server.html)...

integreat_compass/cms/forms/custom_model_form.py
timobrembeck: Good point, the integreat cms also carries along quite a bit of technical debt, so if you find an improvement of the existing base classes, feel free to open a PR in the integreat cms repo as well to backport the refactored solution...

integreat_compass/cms/views/offers/offer_form_view.py
timobrembeck: Also, the printed object itself is meaningless since you are in the `else` case of `all(form.is_valid())`, so we can be 100% sure that `contact_form.is_valid()` returns `True` at this point in time.

integreat_compass/cms/views/offers/offer_form_view.py
timobrembeck: To my knowledge, the `commit=False` flag already is the "django-way" of handling this use case, see [docs](https://docs.djangoproject.com/en/4.2/topics/forms/modelforms/#the-save-method).

integreat_compass/cms/forms/custom_model_form.py
```diff
@@ -0,0 +1,188 @@
...
+        instance = super().save(commit=False)
+        if hasattr(self.instance, "user"):
+            instance.creator = self.instance.user
```
timobrembeck: What exactly is the reason for the different names of the attributes?
Why set `instance.user = request.user` and then later `instance.creator = instance.user`? Why not directly `instance.creator = request.user` via `additional_instance_attributes`?

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+        offer_instance = Offer.objects.filter(id=kwargs.get("offer_id")).first()
+        if not request.user.has_perm("cms.edit_offer", offer_instance):
+            raise PermissionDenied(
+                f"{request.user!r} does not have the permission to edit {offer_instance!r}"
+            )
+
+        user = User.objects.get(pk=request.user.id)
+
+        offer_form = OfferForm(
+            data=request.POST, additional_instance_attributes={"user": user}
```
timobrembeck:
```suggestion
            data=request.POST, additional_instance_attributes={"creator": user}
```
I think we could directly set the creator here? Then we wouldn't need the custom `´creator = user` statement in the `save()` method`?

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        offer_version_form = OfferVersionForm(
+            data=request.POST, additional_instance_attributes={"user": user}
```
timobrembeck:
```suggestion
            data=request.POST
```
The offer neither has a `creator`, nor a user `field`, so why is the additional attribute set?

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        contact_form = ContactForm(
+            data=request.POST, additional_instance_attributes={"user": user}
```
timobrembeck:
```suggestion
            data=request.POST, additional_instance_attributes={"creator": user}
```
Here maybe also set the creator directly?

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        location_form = LocationForm(
+            data=request.POST, additional_instance_attributes={"user": user}
```
timobrembeck:
```suggestion
            data=request.POST
```
The `Location` model does not have a `creator` or `user`

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        organization_form = OrganizationForm(
+            data=request.POST, additional_instance_attributes={"user": user}
```
timobrembeck:
```suggestion
            data=request.POST, additional_instance_attributes={"creator": user}
```

integreat_compass/cms/apps.py
```diff
@@ -19,4 +19,4 @@ def ready(self):
         # pylint: disable-next=import-outside-toplevel
         from django.contrib.auth.models import Group
 
-        Group.__str__ = lambda self: _(self.name)
+        Group.__str__ = lambda self: self.name
```
timobrembeck: If you remove the translation, you can also remove the whole line.
The default string representation is already the name:
https://github.com/django/django/blob/c3862735cd8c268e99fb8d54c3955aacc4f2dc25/django/contrib/auth/models.py#L129-L130

integreat_compass/core/middleware/access_control_middleware.py
```diff
@@ -40,6 +40,7 @@ def __call__(self, request):
         if (
             namespaces.PUBLIC not in resolver_match.namespaces
             and not request.user.is_authenticated
+            and not "admin" in request.path
```
timobrembeck:
```suggestion
            and "admin" not in resolver_match.namespaces
```
We don't want this to happen for e.g. `/offers/badminton-course/` :joy:

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+        :raises ~django.core.exceptions.PermissionDenied: If user does not have the permission to edit the specific page
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        offer_instance = Offer.objects.filter(id=kwargs.get("offer_id")).first()
+        if not request.user.has_perm("cms.edit_offer", offer_instance):
+            raise PermissionDenied(
+                f"{request.user!r} does not have the permission to edit {offer_instance!r}"
+            )
```
timobrembeck:
```suggestion
```
Since the `dispatch()` method is already decorated with `@method_decorator(permission_required("cms.edit_offer"), name="dispatch")`, this case will never be hit, since `post()` is not called in this case anyway.

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+                offer.organization,
+            )
+            if offer is not None
+            else (None, None, None, None)
+        )
+        offer_form = OfferForm(instance=offer)
+        offer_version_form = OfferVersionForm(instance=offer_version)
+        contact_form = ContactForm(instance=offer_contact)
+        location_form = LocationForm(instance=offer_location)
+        organization_form = OrganizationForm(instance=offer_organization)
```
timobrembeck: You could also outsource the form init functions into a separate function that's called from both `get()` and `post()` to remove code duplication.
Django does it is a similar way in the [FormMixin](https://github.com/django/django/blob/c3862735cd8c268e99fb8d54c3955aacc4f2dc25/django/views/generic/edit.py#L29C8-L53).

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+    OfferVersionForm,
+    OrganizationForm,
+)
+from ...models import Offer, User
+
+logger = logging.getLogger(__name__)
+
+
+@method_decorator(permission_required("cms.edit_offer"), name="dispatch")
+class OfferFormView(TemplateView):
```
timobrembeck, line 24:
```suggestion
class OfferFormView(SingleObjectMixin, TemplateView):
```
You could let this view inherit from [SingleObjectMixin](https://docs.djangoproject.com/en/4.2/ref/class-based-views/mixins-single-object/#singleobjectmixin) and if you want to make use of the inbuilt `get_object()` method to retrieve the current offer instance. This works best if the url parameter is renamed from `offer_id` to just `pk`.

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,178 @@
...
+        :raises ~django.core.exceptions.PermissionDenied: If user does not have the permission to edit the specific page
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+
+        offer = Offer.objects.filter(id=kwargs.get("offer_id")).first()
+        (offer_version, offer_contact, offer_location, offer_organization) = (
+            (
+                offer.versions.latest(),
```
timobrembeck: Since the `Offer` model does not have a `get_latest_by` meta attribute, `latest()` does not work without providing a field name. Did you mean `first()`?

ztefanie:
Please add tests for this.

integreat_compass/cms/models/offers/tag.py
charludo: Both the "name" fields of contact and organization, as well as the "title" fields of tag and offer_version were clashing, since all of those values are stored in a single request object when submitting the form.

integreat_compass/cms/models/organizations/contact.py
charludo: See above

integreat_compass/cms/views/offers/offer_form_view.py
charludo: Sorry, this was just for my own debugging purposes and should already have been removed.

Used print because currently it didn't  make a difference for that purpose since coloring of the logs seems broken at the moment.

integreat_compass/cms/views/offers/offer_form_view.py
charludo: I think on its own this way is already rather concise, it's just that we have to do it for multiple forms which results in annoyingly many lines.

integreat_compass/cms/forms/custom_model_form.py
charludo: > `additional_instance_attributes`

that was my initial idea as well, but I could not get this to work correctly.

I think it's just that there is no instance initially.

integreat_compass/cms/views/offers/offer_form_view.py
charludo: See above

integreat_compass/cms/apps.py
charludo: Alright. Just for an explanation: This translation prevented the creation of new groups in the backend, throwing `TypeError: __str__ returned non-string (type __proxy__)`. See also: https://stackoverflow.com/questions/51959209/django-error-str-returned-non-string-type-proxy

Using the non-lazy translate version gave a different set of errors, so I just opted to removing it.

integreat_compass/core/middleware/access_control_middleware.py
charludo: Hahaha, good point 😂

integreat_compass/cms/models/offers/tag.py
ztefanie: I think what you should use is: https://docs.djangoproject.com/en/4.2/ref/forms/api/#prefixes-for-forms
Please undo this renamings, it should definately not be necessary to have unique attribute names across different models. Maybe this article helps, too: https://collingrady.wordpress.com/2008/02/18/editing-multiple-objects-in-django-with-newforms/

charludo: Neat, did not know about the prefixes. That looks like a good idea, thanks.

integreat_compass/cms/forms/custom_model_form.py
timobrembeck: In theory, the base class should automatically create an instance object if it doesn't really exist. The only difference of the new instance is that it does not yet have an id, but setting attributes should work even before it's saved to the database... or is the problem that the creator is a related field that cannot be assigned before saving? 🤔
And don't we use the same pattern for integreat-cms as well?

charludo: I'll have another go at it, maybe the issue lay elsewhere. But neither `additional_instance_attributes` nor setting the field during `clean` worked, the only place I got it working was after saving without committing, so I think you might be onto something re: related field.

If you do  not think it makes sense to have this in the `CustomModelForm` though, we could of course just do it in the `OfferFormView`.

integreat_compass/cms/models/interactions/comment.py
charludo: Two options what we could do here:
- `SET_NULL`: would make the most sense, in our templates we could simply show something like `[deleted user]` as their name if the reference is null. However, this requires creator to be nullable which I don't really like...
- `SET()`: we could also set a "fallback" user here... Avoids the nullable field but feels hacky.

timobrembeck: In my opinion, the first option with a nullable field and `SET_NULL` makes more sense :+1:

timobrembeck: Alternatively, you could also keep old users for reference, and only disable their login by setting `is_active=False`...
Then, you could use `PROTECT` instead of `CASCADE` or `SETNULL`.

charludo: From a techincal perspective yes, but since the users do have a right to be forgotten, we'd have to at least also remove their email and display name, at which point, deleting them properly might make more sense? 😅

integreat_compass/cms/forms/custom_model_form.py
charludo: > additional_instance_attributes

No idea what caused my previos issue, now it works as intended 😅

integreat_compass/cms/views/offers/offer_form_view.py
charludo: If I'm not missing anything, this actually makes the code more verbose - if no pk or slug are passed in the url, `get_object()` throws an error. Is it worth using and catching this?

integreat_compass/cms/views/offers/offer_form_view.py
charludo: Could you elaborate on this? 😅  Because from what I'm seeing, each form in `get` requires a unique argument, and different arguments yet again in `post`.

charludo: I mean, I could add something like `_get_get_forms()` and `_get_post_forms()` of course, which would make the existing methods shorter, but does not really reduce code duplication 😅

integreat_compass/cms/views/offers/offer_form_view.py
timobrembeck: Ah ok, right, we'd need a base class for both the update & create views (e.g. `OfferFormMixin`), and then two views which inherit from this (e.g. `CreateOfferView` with `self.object = None` and `UpdateOfferView` with `self.object = self.get_object()`) and instantiate the corresponding view in the URL config, at least this is how Django is doing it:
https://github.com/django/django/blob/c3862735cd8c268e99fb8d54c3955aacc4f2dc25/django/views/generic/edit.py#L169-L206

Could also be more work than necessary, so I'm also fine with keeping it the way it is.
I'd just prefer if we could throw an error if an invalid id is given (at the moment, this is equivalent to creating a new offer).

integreat_compass/cms/views/offers/offer_form_view.py
timobrembeck: I didn't try this out, but imagined something along the lines of:
```
    def get_form_classes(self):
        return [ContactForm, LocationForm, OfferForm, OfferVersionForm, OrganizationForm]

    def get_form_kwargs(self, form_class):
        kwargs = {}
        if self.request.method in ["POST", "PUT"]:
            kwargs["data"] = self.request.POST
        if hasattr(self, "object"):
            kwargs["instance"] = form_class.get_instance_from_offer(self.object)
        return kwargs

    def get_forms(self):
        return [
            form_class(**self.get_form_kwargs(form_class))
            for form_class in self.get_form_classes()
        ]

    def get_context_data(self, **kwargs):
        for form in self.get_forms():
            kwargs[form.context_name] = form
        return super().get_context_data(**kwargs)
```

This would require to implement a `get_instance_from_offer()` method in each form class, and an attribute `context_name` which denotes the name of the variable to access the form in the template...

Not sure however if this really makes things easier or just more complicated :sweat_smile:

charludo: It's a nice idea, but introducing (essentially) 9 new methods to avoid 5 lines not-quite-duplicate lines of code is probably excessive 😅

integreat_compass/cms/views/offers/offer_form_view.py
charludo: The permission check and loading the user are gone now. I don't really see a reason to split up the remaining method - all we are doing is getting and saving the forms. It makes sense to do this in one method IMO.

It's just that we have so many forms that this produces numerous lines, but since this can't really be generalized into a (general) helper method, I think keeping it as it is now is the most readable version.

integreat_compass/cms/views/offers/offer_form_view.py
timobrembeck: Yes you're right :see_no_evil: 
Maybe if we happen to add more views with multiple forms :grin:

integreat_compass/cms/views/offers/offer_form_view.py
timobrembeck: Ah, but is there really a reason for `commit=False` in the first place?
Why not just do
```
            offer_form.instance.offer_contact = contact_form.save()
            offer_form.instance.location = location_form.save()
            offer_form.instance.organization = organization_form.save()
            offer_form.save()
```

instead of 
```
            offer_instance = offer_form.save(commit=False)
            offer_instance.offer_contact = contact_form.save()
            offer_instance.location = location_form.save()
            offer_instance.organization = organization_form.save()
            offer_instance.save()
            offer_form.save_m2m()
```
?

In think in my comment above I overlooked the fact that only many-to-many relationships require the object to be saved before the relationship can be assigned, for simple foreign keys you can just assign the fields of an unsaved model instance.

charludo: Yep, that works as well! Pushed the change.

integreat_compass/cms/forms/custom_model_form.py
```diff
@@ -0,0 +1,179 @@
...
+            "%s initialized"
+            + (" with " + ": %r, ".join(attributes) + ": %r" if attributes else ""),
+            type(self).__name__,
+            *[getattr(self, attribute) for attribute in attributes],
+        )
+
+    def _init_fields(self, disabled):
+        for field_name, field in self.fields.items():
+            if disabled:
+                field.disabled = True
```
timobrembeck:
```suggestion
            field.disabled = disabled
```

integreat_compass/cms/forms/custom_model_form.py
```diff
@@ -0,0 +1,179 @@
...
+        :param commit: Whether or not the changes should be written to the database
+        :type commit: bool
+
+        :return: The saved object returned by :ref:`django:topics-modelform-save`
+        :rtype: object
+        """
+        self.logger.debug(
+            "%s saved with changed data %r", type(self).__name__, self.changed_data
+        )
+        self.data = self.cleaned_data
```
timobrembeck:
```suggestion
```
This does not really make sense. The `data` contains the raw data submitted via the form (as strings), the `cleaned_data` contains the data converted to their corresponding Python representation (so e.g. model objects)... mixing those will have weird side effects.

integreat_compass/cms/forms/custom_model_form.py
```diff
@@ -0,0 +1,179 @@
...
+
+        instance = super().save(commit=commit)
+        return instance
```
timobrembeck:
```suggestion
        return super().save(commit=commit)
```

integreat_compass/cms/forms/offers/offer_version_form.py
```diff
@@ -0,0 +1,26 @@
...
+
+    class Meta:
+        """
+        This class contains additional meta configuration of the form class, see the :class:`django.forms.ModelForm`
+        for more information.
+        """
+
+        model = OfferVersion
+        fields = ["title", "description", "language", "is_free"]
+        widgets = {"is_free": forms.Select(choices=((True, "Yes"), (False, "No")))}
```
timobrembeck:
```suggestion
        widgets = {"is_free": forms.Select(choices=((True, _("Yes")), (False, _("No"))))}
```
Probably makes sense to translate these? (requires the import)

integreat_compass/cms/models/offers/language.py
```diff
@@ -28,7 +28,7 @@ def __str__(self):
         :return: A readable string representation of the page
         :rtype: str
         """
-        return f"Language: {self.english_name}"
+        return self.english_name
```
timobrembeck:
```suggestion
        return _(self.english_name)
```
Why not translate it?

integreat_compass/cms/templates/offers/offer_form.html
```diff
@@ -0,0 +1,132 @@
+{% extends "_base.html" %}
+{% load i18n %}
+{% load static %}
+{% load widget_tweaks %}
+{% load render_bundle from webpack_loader %}
+{% block title %}
+    {% if offer %}
+        {% blocktranslate trimmed %}
+            Edit offer "{{ offer.public_version.title }}"
```
timobrembeck:
```suggestion
    {% if offer_version_form.instance.id %}
        {% blocktranslate trimmed with offer_title=offer_version_form.instance.title %}
            Edit offer "{{ offer_title }}"
```

integreat_compass/cms/templates/offers/offer_form.html
```diff
@@ -0,0 +1,132 @@
...
+    <form enctype="multipart/form-data"
+          method="post"
+          id="offer_form"
+          data-unsaved-warning>
+        {% csrf_token %}
+        <div class="w-full flex flex-wrap justify-between mb-6 gap-2">
+            <h1 class="heading">
+                {% if offer %}
+                    {% blocktranslate trimmed %}
+                        Edit offer "{{ offer.public_version.title }}"
```
timobrembeck:
```suggestion
                {% if offer_version_form.instance.id %}
                    {% blocktranslate trimmed with offer_title=offer_version_form.instance.title %}
                        Edit offer "{{ offer_title }}"
```

integreat_compass/cms/apps.py
timobrembeck: What errors did the non-lazy version return?
You just have to make sure that you import both variants, the lazy version for the `verbose_name` of the config, and the non-lazy one for the group str...

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,158 @@
...
+        (offer_version, offer_contact, offer_location, offer_organization) = (
+            (
+                offer.versions.first(),
+                offer.offer_contact,
+                offer.location,
+                offer.organization,
+            )
+            if offer is not None
+            else (None, None, None, None)
+        )
```
timobrembeck:
```suggestion
        if offer_id := kwargs.get("pk"):
            offer = Offer.objects.get(id=offer_id)
            offer_version = offer.versions.first()
            offer_contact = offer.offer_contact
            offer_location = offer.location
            offer_organization = offer.organization
        else:
            offer = offer_version = offer_contact = offer_location = offer_organization = None  # fmt: skip
```
Maybe this would be more concise?

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,158 @@
...
+            data=request.POST, additional_instance_attributes={"creator": request.user}
+        )
+        offer_version_form = OfferVersionForm(data=request.POST)
+        contact_form = ContactForm(
+            data=request.POST, additional_instance_attributes={"creator": request.user}
+        )
+        location_form = LocationForm(data=request.POST)
+        organization_form = OrganizationForm(
+            data=request.POST, additional_instance_attributes={"creator": request.user}
+        )
```
timobrembeck:
```suggestion
        if offer_id := kwargs.get("pk"):
            offer = Offer.objects.get(id=offer_id)
            offer_version = offer.versions.first()
            offer_contact = offer.offer_contact
            offer_location = offer.location
            offer_organization = offer.organization
        else:
            offer = offer_version = offer_contact = offer_location = offer_organization = None  # fmt: skip

        form_kwargs = {
            "data": request.POST,
            "additional_instance_attributes": {"creator": request.user},
        }

        offer_form = OfferForm(instance=offer, **form_kwargs)
        offer_version_form = OfferVersionForm(instance=offer_version, **form_kwargs)
        contact_form = ContactForm(instance=offer_contact, **form_kwargs)
        location_form = LocationForm(instance=offer_location, **form_kwargs)
        organization_form = OrganizationForm(instance=offer_organization, **form_kwargs)
```
You need to pass the `instance` for POST requests as well, otherwise you will just create new instances instead of editing existing ones. At this point, we can again think about de-duplicating the initialization process :laughing:

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,158 @@
...
+            offer_version_form.instance.offer = offer_form.save()
+            offer_version_instance = offer_version_form.save()
+
+            messages.success(
+                request,
+                _(
+                    'Application for the offer "{}" has successfully been submitted and will be reviewed shortly.'
+                ).format(offer_version_instance.title),
+            )
+
```
timobrembeck, line 138:
```suggestion
            return redirect(
                "cms:protected:edit_offer", **{"pk": offer_form.instance.id}
            )

```
Typically, you want to redirect after a successful form submission, otherwise the same data might get submitted again when the page is refreshed...

integreat_compass/cms/templates/offers/offer_form.html
```diff
@@ -0,0 +1,132 @@
...
+                        {% for field in organization_form %}
+                            <div>
+                                <label for="{{ field.id_for_label }}"
+                                       {% if field.field.required %}class="field-required"{% endif %}>
+                                    {{ field.label }}
+                                </label>
+                                {% render_field field %}
+                                <div class="help-text">{{ field.help_text }}</div>
+                            </div>
+                        {% endfor %}
```
timobrembeck, line 125:
```suggestion
                        {{ organization_form }}
```
If you're not rendering the fields individually, you can also go with the default Django rendering... this works for the other forms as well.
(just rename the `help-text` class to `helptext` which is the default class assigned to it)

integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -0,0 +1,158 @@
...
+    OfferForm,
+    OfferVersionForm,
+    OrganizationForm,
+)
+from ...models import Offer
+
+logger = logging.getLogger(__name__)
+
+
+@method_decorator(permission_required("cms.edit_offer"), name="dispatch")
```
timobrembeck:
```suggestion
@method_decorator(permission_required("cms.change_offer"), name="dispatch")
```
I think we could also remove the custom permission `edit_offer`, since it essentially means the same thing as the default permission `change_offer`, right?

integreat_compass/cms/fixtures/permissions.json
```diff
@@ -24,9 +24,8 @@
     "fields": {
       "name": "OFFER_PROVIDER",
       "permissions": [
-        ["edit_offer", "cms", "offer"],
-        ["edit_offer_version", "cms", "offerversion"],
-        ["upload_document", "cms", "document"]
+        ["upload_document", "cms", "document"],
+        ["edit_offer_version", "cms", "offerversion"]
```
timobrembeck:
```suggestion
        ["change_offerversion", "cms", "offerversion"]
```
Also there I think there is no need to duplicate the existing default permission?

integreat_compass/cms/models/offers/offer_version.py
```diff
@@ -103,6 +103,6 @@ class Meta:
         verbose_name = _("offer version")
         verbose_name_plural = _("offer versions")
         default_related_name = "offer_versions"
-        ordering = ["offer_version_date"]
+        ordering = ["-offer_version_date"]
         default_permissions = ("change", "delete", "view")
         permissions = [("edit_offer_version", "Can change offer version details")]
```
```suggestion
```
This can probably removed since we have the default `change` permission?

integreat_compass/cms/templates/_base.html
```diff
@@ -15,18 +15,52 @@
             <a href="#" class="ml-6 mr-4 hover:text-gray-900">{% translate "All offers" %}</a>
             {% if perms.cms.edit_offer %}
```
timobrembeck:
```suggestion
            {% if perms.cms.change_offer %}
```

timobrembeck (changes requested):

integreat_compass/cms/forms/custom_model_form.py
charludo: 🤦🏼‍♀️ 
Not sure if the fact that this is copied from integreat makes it better or worse 😅

integreat_compass/cms/forms/offers/offer_version_form.py
charludo: Ah, yes! Sorry, forgot to add.

integreat_compass/cms/views/offers/offer_form_view.py
charludo: Yes, though I think I'd rather redirect to the user's offer overview, but since this did not exist yet, the placeholder.

integreat_compass/cms/forms/custom_model_form.py
timobrembeck: Haha, I don't want to claim that the code I have written over the past years cannot be improved :grin:
In this specific case, it's mostly personal preference, so if you prefer the current state, I'm completely fine with it.

integreat_compass/cms/fixtures/permissions.json
charludo: TBH I did not have a close look at this, these are dumped from the DB without manual edits.

integreat_compass/cms/apps.py
charludo: `AppRegistryNotReady` - but yes, with the lazy for the verbose_name it appears to be working again 🤔

integreat_compass/cms/views/offers/offer_form_view.py
charludo: Is i? 😅 But sure, I do not mind either way

integreat_compass/cms/templates/offers/offer_form.html
charludo: Will this use the `WIDGET_ERROR_CLASS` though? I think that is widget-tweaks specific, right? Also the additional class on the label fir required fields.

I know it is a bit more verbose, but it's probably easier to keep everything consistent?

integreat_compass/cms/views/offers/offer_form_view.py
charludo: Done, `get` can't get much more concise now... 😄

integreat_compass/cms/templates/offers/offer_form.html
timobrembeck: Ah, right :+1:

integreat_compass/cms/urls/public.py
```diff
@@ -1,5 +1,14 @@
...
 
-urlpatterns = [path("", index.IndexView.as_view(), name="index")]
+urlpatterns = [
+    path("", index.IndexView.as_view(), name="index"),
+    path(
+        "accounts/login/",
+        auth_views.LoginView.as_view(template_name="login.html"),
+        name="login",
+    ),
+    path("accounts/logout/", auth_views.LogoutView.as_view(), name="logout"),
```
timobrembeck:
```suggestion
        "accounts/",
        include(
            [
                path(
                    "login/",
                    auth_views.LoginView.as_view(template_name="login.html"),
                    name="login",
                ),
                path("logout/", auth_views.LogoutView.as_view(), name="logout"),
            ]
        ),
    ),
```

integreat_compass/cms/urls/public.py
```diff
@@ -1,5 +1,14 @@
+from django.contrib.auth import views as auth_views
 from django.urls import path
```
timobrembeck:
```suggestion
from django.urls import include, path
```

integreat_compass/cms/templates/index.html
```diff
@@ -8,7 +8,7 @@
         <div class="flex flex-col justify-center gap-4 text-center">
             <h1 class="font-bold text-5xl drop-shadow-md mb-12 text-white">{% translate "Arriving, reimagined." %}</h1>
             <a href="#"
-               class="bg-redMuted hover:bg-red-500 text-white font-bold text-3xl py-4 px-4 rounded-2xl">{% translate "Discover Integreat Compass" %}</a>
+               class="bg-primary hover:bg-red-500 text-white font-bold text-3xl py-4 px-4 rounded-2xl">{% translate "Discover Integreat Compass" %}</a>
             <a href="#"
```
timobrembeck:
```suggestion
            <a href="{% url 'cms:protected:new_offer' %}"
```

timobrembeck (changes requested):
Very cool, thanks! :muscle: 
The form view looks pretty clean now!

I also think it would be cool to add a few tests, just a few ideas for test cases:
- The offer form view correctly returns a redirect to the login when accessed by an anonymous user
- The offer form view correctly returns error 403 if the user is logged in but does not have the permission `change_offer`
- The offer form does not throw an exception when the form is rendered (both the `new` form and `edit`)
- The offer form view correctly creates db objects when submitted to `new`
- The offer form view correctly updates existing db objects when submitted to `edit`

README.md
```diff
@@ -30,7 +30,15 @@ cd integreat-compass
...
 * Go to your browser and open the URL `http://localhost:8082`
-* Default user is "root" with password "root1234".
+* By default, the following users exist:
+
+| Email                                    | Password | Group                  |
+|------------------------------------------|----------|------------------------|
+| root@integreat.compass                   | root1234 | -                      |
+| board_member@integreat.compass           | bm1234   | BOARD_MEMBER           |
+| integration_specialist@integreat.compass | is1234   | INTEGRATION_SPECIALIST |
+| offer_provider@integreat.compass         | op1234   | OFFER_PROVIDER         |
```
timobrembeck: If I had to find something to complain about, it would be that these test passwords are a bit hard to remember, and since they're public anyway, I don't see a reason why we wouldn't just use `test` or `password` as password for all of them (Django also allows to override its security requirements when the user is created via the management command).
But ok, they will probably be saved to the browser's password manager anyways^^

tests/cms/test_login.py
timobrembeck, line 1: Just for the future: Whenever you use Django's inbuilt views (like `LoginView` and `LogoutView` in this case), you can rely on the fact that they have [their own testing framework](https://github.com/django/django/blob/d3d173425fc0a1107836da5b4567f1c88253191b/tests/auth_tests/test_views.py), so we don't need to test the same things again.

tests/conftest.py
```diff
@@ -23,3 +29,76 @@ def load_test_data(django_db_setup, django_db_blocker):
...
+    :return: The http client and the current role
+    :rtype: tuple
+    """
+    async_client = AsyncClient()
+    # Only log in user if the role is not anonymous
+    if request.param != ANONYMOUS:
+        with django_db_blocker.unblock():
+            user = get_user_model().objects.get(username=request.param.lower())
+            async_client.force_login(user)
+    return async_client, request.param
```
```suggestion
```
async tests cause a lot of side-effects and potential problems, so I wouldn't use them if not strictly necessary. For this project, the synchronous tests are hopefully sufficient...

timobrembeck (approved):
Awesome :tada:
Just a few optional nit-picks, but in general this PR is ready to be merged imo! :+1:

README.md
charludo: Hm... Fair point. I wanted to continue the `root`=> `root1234` pattern, but the other role names were too long for my taste... 😂

tests/cms/test_login.py
charludo: Sure, but doesn't it still make some sense to test the `LoginView`, since we've customized *some* behaviour (email instead of username, custom template)?

README.md
timobrembeck: I guess the only reason I appended `1234` to the root password for integreat-cms was that the system complained that `root` was too insecure and I didn't know how to override the warning at that time :joy:

tests/cms/test_login.py
timobrembeck: oh right sorry, yeah I missed that :+1:

README.md
charludo: I could just change all the passwords to e.g. `compass`, if you'd prefer, and edit the readme/docs accordingly. I'll leave it up to you.

timobrembeck: would be cool, thanks :sweat_smile:

ztefanie (approved):

Issue #36, opened 2023-04-20
Add production deployment workflow to CircleCI


timobrembeck:
### Motivation

Similar to #11, but for production deployments and not dev deployments.
This workflow can be triggered by merging a PR into the main branch.

### Proposed Solution

- Add a workflow `main` that runs on each commit on main
  - Add a job that bumps the version and commits the changes to the main branch and creates a git tag (can be inspired by [bump-version](https://github.com/digitalfabrik/integreat-cms/blob/9d7c4da6624949d39d404e8977b4346aec87943a/.circleci/config.yml#L433-L482) job of integreat-cms)
  - Merge version bump back into develop branch
- Add a workflow `deploy` that runs on each new git tag
  - Add jobs for all requirements for the production package, e.g.:
    - Compile translations
    - Bundle static files via webpack
  - Add job to build the main package with `python3 -m build`
  - Add job to publish the package to PyPI via twine (see [publish-package](https://github.com/digitalfabrik/integreat-cms/blob/9d7c4da6624949d39d404e8977b4346aec87943a/.circleci/config.yml#L322-L335) job of integreat-cms)
  - Add job to create GitHub release (see [create-release](https://github.com/digitalfabrik/integreat-cms/blob/9d7c4da6624949d39d404e8977b4346aec87943a/.circleci/config.yml#L483-L517) job of integreat-cms)
  - Add job to notify mattermost about the release (see [notify-mattermost](https://github.com/digitalfabrik/integreat-cms/blob/9d7c4da6624949d39d404e8977b4346aec87943a/.circleci/config.yml#L518-L547) job of integreat-cms)
 
### Alternatives

Manual deployments

This depends on:
- #10

PR #57, opened 2023-05-29
Add publishing tool

### LINKED ISSUE #11: Add dev package publishing job to CircleCI workflow
Each commit on the develop branch should automatically:
- [bump the dev version](https://github.com/digitalfabrik/integreat-cms/blob/68a3810c3b20cc83d1fb93d71e6ed5af60f591dc/.circleci/config.yml#L398-L432) to a temporary beta version (this does not need to be committed to the develop branch, it's just to allow new versions to be pushed to pypi)
- [trigger a new package build](https://github.com/digitalfabrik/integreat-cms/blob/68a3810c3b20cc83d1fb93d71e6ed5af60f591dc/.circleci/config.yml#L301-L321) with `python3 -m build`
- [publish the new package](https://github.com/digitalfabrik/integreat-cms/blob/68a3810c3b20cc83d1fb93d71e6ed5af60f591dc/.circleci/config.yml#L322-L335) on https://test.pypi.org/ via twine

This depends on:
- #10

JoeyStk:
### Short description
Add dev package publishing job to CircleCI workflow

### Proposed changes
- bump the dev version to a temporary beta version
- trigger a new package build with python3 -m build
- publish the new packageon https://test.pypi.org/ via twine

### Side effects

- I hope none

### Resolved issues

Fixes: #11

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,29 @@ jobs:
       - run:
           name: Check black code style
           command: black --check .
+  compile-translations:
+    docker:
+      - image: cimg/python:3.9.16
+    resource_class: small
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Install gettext
+          command: sudo apt-get update && sudo apt-get install gettext
+      - run:
+          name: Enable virtual environment
+          command: echo "source .venv/bin/activate" >> $BASH_ENV
+      - run:
+          name: Compile translation file
+          command: |
+            cd integreat_compass
+            integreat-compass-cli compilemessages --settings=integreat_compass.core.circleci_settings
+      - persist_to_workspace:
+          root: .
+          paths:
+            - integreat_compass/locale/*/LC_MESSAGES/django.mo
   djlint:
     docker:
       - image: cimg/python:3.9.16
@@ -133,6 +156,94 @@ jobs:
       - run:
           name: Run pylint
           command: ./tools/pylint.sh
+  webpack:
+    docker:
+      - image: "cimg/node:lts"
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Compile and bundle CSS and JS
+          command: npm run prod
+      - persist_to_workspace:
+          root: .
+          paths:
+            - integreat_compass/static/dist
+            - integreat_compass/webpack-stats.json
+  build-package:
+    docker:
+      - image: cimg/python:3.9.16
+    resource_class: small
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Enable virtual environment
+          command: echo "source .venv/bin/activate" >> $BASH_ENV
+      - run:
+          name: Build integreat-compass package
+          command: python3 -m build
+      - persist_to_workspace:
+          root: .
+          paths:
+            - dist
+  bump-dev-version:
+    docker:
+      - image: cimg/python:3.9.16
+    resource_class: small
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Enable virtual environment
+          command: echo "source .venv/bin/activate" >> $BASH_ENV
+      - run:
+          name: Bump version
+          command: |
+            # Install recent version of pip
+            echo "Upgrade pip to make sure 'pip index' is available"
+            pip install --upgrade pip
+            # Check which versions of integreat-compass are available on the TestPyPI repository
+            AVAILABLE_VERSIONS=$(pip index versions integreat-compass --pre -i https://test.pypi.org/simple/)
+            echo "Current available versions on TestPyPI: ${AVAILABLE_VERSIONS}"
+            CURRENT_ALPHA_VERSION=$(echo "${AVAILABLE_VERSIONS}" | head -n 1)
+            echo "Most recent version on TestPyPI: ${CURRENT_ALPHA_VERSION}"
+            CURRENT_ALPHA_VERSION=$(echo "${CURRENT_ALPHA_VERSION}" | sed "s/integreat-compass (\([^()]*\)a0)/\1-alpha/")
+            echo "Version converted to alternative format: ${CURRENT_ALPHA_VERSION}"
+            # Get current prod version
+            CURRENT_VERSION=$(python -c "import integreat_compass; print(integreat_compass.__version__)")
+            echo "Current production version: ${CURRENT_VERSION}"
+            # Bump version to current alpha version if it is newer
+            # Attention: exit(True) in Python means exit(1) which is False in Bash :)
+            if python -c "from looseversion import LooseVersion; exit(LooseVersion('${CURRENT_VERSION}') > LooseVersion('${CURRENT_ALPHA_VERSION}'))"; then
+              echo "Bump to the currently existing version"
+              bumpver update -n --set-version="${CURRENT_ALPHA_VERSION}" --no-commit
+            fi
+            # Bump version to next alpha version
+            echo "Bump to the next version"
+            bumpver update -n -t alpha --no-commit
+      - persist_to_workspace:
+          root: .
+          paths:
+            - pyproject.toml
+            - integreat_compass/__init__.py
+  publish-package:
+    docker:
+      - image: cimg/python:3.9.16
+    resource_class: small
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Enable virtual environment
+          command: echo "source .venv/bin/activate" >> $BASH_ENV
+      - run:
+          name: Publish integreat-compass package to (Test-)PyPI
+          command: twine upload --non-interactive ./dist/integreat-compass-*.tar.gz
 
 workflows:
   develop:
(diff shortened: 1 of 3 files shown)


timobrembeck (changes requested):
There have been a few bug fixes on the develop branch of integreat-cms since the lines that you copied. Could you checkout the recent state and update the job?
Also, the new jobs are not referenced in any workflows, which means they are not executed at the moment. Could you add them as well?

.circleci/config.yml
```diff
@@ -169,13 +245,42 @@ workflows:
...
+              only:
+                - develop
+                - /.*-publish-dev-package/
+          requires:
+            - pip-install
+      - build-package:
+          name: build-dev-package
+          requires:
+            # - webpack
+            # - compile-translations
```
JoeyStk: I wasn't sure about these two lines. As far as I understand these lines only apply to integreat-cms and not to the compass, right? If so, I'll remove them

timobrembeck: Since we also use webpack and translations for this project, we also need the corresponding jobs here.
We don't want webpack to be running on the production server, thus we need one step to bundle the static files and put them in the python package as well.
Same with the translations, since we don't have the compiled translations in our repository, we need to compile them first and pass them to the packaging job in order for them to be present in the resulting package.

JoeyStk: Ok, thank you for the explanation. It's still going to take a minute to wrap my head around it, but this helps for now. Is it enough for this PR to comment the two jobs in again or is it also part of this PR to define the webpack job [like done here](https://github.com/digitalfabrik/integreat-cms/blob/develop/.circleci/config.yml#L197-L211)?

timobrembeck: We have to add the job definition here, otherwise the workflow will just fail because the required jobs do not exist.

.circleci/config.yml
```diff
@@ -72,6 +72,29 @@ jobs:
...
+          name: Install gettext
+          command: sudo apt-get update && sudo apt-get install gettext
+      - run:
+          name: Enable virtual environment
+          command: echo "source .venv/bin/activate" >> $BASH_ENV
+      - run:
+          name: Compile translation file
+          command: |
+            cd integreat_compass
+            integreat-compass-cli compilemessages --settings=integreat_compass.core.circleci_settings
```
JoeyStk, line 93: I have another question, this time about this line. So this lines refers to the circleci_settings.py file, which this far doesn't exist. However up until now the file mainly just imports the settings.py file. Therefore I would propose to reference the settings file in this line. What do you think?

timobrembeck: I think we need the dedicated circleci settings file to set a dummy secret key, also it's helpful because we can enable DEBUG logging on CircleCI.

JoeyStk: Ok, thank you. I already prepared the file like that. But I had a second question about it. Is there a way to circumvent the wildcard import at the beginning? As far as I understand we need all and not just a few dedicated imports from the settings, but also as far as I know this isn't really good practice. So now I'm wondering, what would be best for me to do?

timobrembeck: Normally, this is bad practice because it imports more than necessary, but right, in this case we want all settings to be present on the custom settings file as well. So there is no possibility to circumvent the wildcard import apart from listing all settings explicitly - but since this are quite a few this doesn't really make sense, also it has the disadvantage that the list needs to be adapted as soon as a setting in the original settings file is added/removed, so in this case the wildcard import is exactly what we want.

JoeyStk: Ok, thank you for clearing that up

.circleci/config.yml
JoeyStk: The requested change should be applied

integreat_compass/core/circleci_settings.py
```diff
@@ -0,0 +1,14 @@
...
+For more information on this file, see :doc:`django:topics/settings`.
+For the full list of settings and their values, see :doc:`django:ref/settings`.
+"""
+# pylint: disable=wildcard-import
+# pylint: disable=unused-wildcard-import
+from .settings import *
+
+#: Set a dummy secret key for CircleCI build even if it's not in debug mode
+SECRET_KEY = "dummy"
+#: Use debug logging on CircleCI
```
ztefanie: Please remove this comment, it adds no info at all

ztefanie (approved):
Looks good

integreat_compass/core/circleci_settings.py
JoeyStk: Is applied now :)

.circleci/config.yml
```diff
@@ -133,6 +156,97 @@ jobs:
...
+            CURRENT_ALPHA_VERSION=$(echo "${AVAILABLE_VERSIONS}" | head -n 1)
+            echo "Most recent version on TestPyPI: ${CURRENT_ALPHA_VERSION}"
+            CURRENT_ALPHA_VERSION=$(echo "${CURRENT_ALPHA_VERSION}" | sed "s/integreat-compass (\([^()]*\)a0)/\1-alpha/")
+            echo "Version converted to alternative format: ${CURRENT_ALPHA_VERSION}"
+            # Get current prod version
+            CURRENT_VERSION=$(python -c "import integreat_compass; print(integreat_compass.__version__)")
+            echo "Current production version: ${CURRENT_VERSION}"
+            # Bump version to current alpha version if it is newer
+            # Attention: exit(True) in Python means exit(1) which is False in Bash :)
+            if python -c "from looseversion import LooseVersion; exit(LooseVersion('${CURRENT_VERSION}') > LooseVersion('${CURRENT_ALPHA_VERSION}'))"; then
```
timobrembeck, line 221: You need to add `looseversion` to the dev dependencies in order for this to work.

.circleci/config.yml
```diff
@@ -133,6 +156,97 @@ jobs:
...
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Enable virtual environment
+          command: echo "source .venv/bin/activate" >> $BASH_ENV
+      - run:
+          name: Publish integreat-compass package to (Test-)PyPI
+          command: twine upload --non-interactive ./dist/integreat-compass-*.tar.gz
```
timobrembeck, line 246: `twine` is required as dev dependency

.circleci/config.yml
```diff
@@ -145,6 +259,20 @@ workflows:
...
+          filters:
+            tags:
+              only: /.*/
+      - compile-translations:
+          name: compile-translations-deploy
+          requires:
+            - pip-install-deploy
+          filters:
+            tags:
+              only: /.*/
```
timobrembeck:
```suggestion
      - webpack:
          requires:
            - npm-install
      - compile-translations:
          requires:
            - pip-install
```
Here, you copied the wrong parts of the integreat-cms config: The `*-deploy` jobs are part of the `deploy` workflow, not the `develop` workflow.

.circleci/config.yml
```diff
@@ -169,13 +297,42 @@ workflows:
...
     jobs:
       - pip-install:
           name: pip-install-main
           filters:
             branches:
               only: main
+      - bump-version:
+          context: deliverino
+          requires:
+            - pip-install-main
```
timobrembeck:
```suggestion
```
You cannot reference a job that doesn't exist yet. You only defined `bump-dev-version`, not `bump-version`.
Also, this PR is only about the develop workflow #11, so you don't need to add jobs for the deploy workflow. The deployment workflow is covered by #36.

.circleci/config.yml
```diff
@@ -185,3 +342,17 @@ workflows:
...
+          filters:
+            tags:
+              only: /.*/
+      - publish-package:
+          context: pypi
+          requires:
+            - build-package
+          filters:
+            tags:
+              only: /.*/
```
timobrembeck:
```suggestion
```
You cannot reference jobs from different workflows as requirements of these jobs.
All jobs that are listed as requirements of jobs in the `deploy` workflow need to be executed in the `deploy` workflow as well.
Also, the same applies as above, so I would just ignore the deployment workflow for now...

timobrembeck (changes requested):

.circleci/config.yml
JoeyStk: Is added now

.circleci/config.yml
JoeyStk: Is also added now

.circleci/config.yml
timobrembeck: Could you please also add the pinned versions of the dependencies?

JoeyStk: looseversion was already included in dev-pinned. I added twine now too

svenseeberg (approved):
IMHO good to go :)

timobrembeck (approved):
Sorry for the delay, yes on first glance I don't see anything blocking this PR either.

timobrembeck:
Ah sorry I forgot: We have to publish one version manually before the tool can start its automatic work.

timobrembeck:
I released an initial version manually, the dev workflow should now work and we can use the test pypi package for setting up a test server:

https://test.pypi.org/project/integreat-compass/

JoeyStk:
> I released an initial version manually, the dev workflow should now work and we can use the test pypi package for setting up a test server

Thank you very much! :blush:

Issue #56, opened 2023-05-26
Revisit model permissions


timobrembeck:
### Motivation

The existing model permissions seem a bit inconsistent to me.
Sometimes, the default permissions are used, and sometimes custom permissions are created for the same task.
I think we should rely on the default permissions wherever possible and remove all custom permissions which essentially mean the same.
Also, we should think about where we want to differentiate between `add` and `change` and where they mean the same.

### Proposed Solution

Some examples:
- `comment_on_offer` seems to be the same as the inbuilt `add_comment`
  https://github.com/digitalfabrik/integreat-compass/blob/bf16fdc3f18042fc171475ed6e555ff7edf58e53/integreat_compass/cms/models/interactions/comment.py#L54
- `set_favorite_offer` is equivalent to `add_favorite`
  https://github.com/digitalfabrik/integreat-compass/blob/bf16fdc3f18042fc171475ed6e555ff7edf58e53/integreat_compass/cms/models/interactions/favorite.py#L43
- `change_favorite` can probably be removed since changing either the `creator` or the `offer` would probably count as `add` & `delete` instead of `change`
  https://github.com/digitalfabrik/integreat-compass/blob/bf16fdc3f18042fc171475ed6e555ff7edf58e53/integreat_compass/cms/models/interactions/favorite.py#L42
- `view_offer_votes` seems to be the same as `view_offervote` and `vote_on_offer` the same as `add_offervote`?
  https://github.com/digitalfabrik/integreat-compass/blob/bf16fdc3f18042fc171475ed6e555ff7edf58e53/integreat_compass/cms/models/interactions/vote.py#L50-L54
- `upload_document` could be considered identical to `add_document`?
  https://github.com/digitalfabrik/integreat-compass/blob/bf16fdc3f18042fc171475ed6e555ff7edf58e53/integreat_compass/cms/models/offers/document.py#LL89C26-L89C41
- `edit_offer` is the same as `change_offer`?
  https://github.com/digitalfabrik/integreat-compass/blob/bf16fdc3f18042fc171475ed6e555ff7edf58e53/integreat_compass/cms/models/offers/offer.py#L91

### Additional information

When a model needs all four default permissions (add, change, delete, view), we can just remove the `default_permissions` attribute from the model's `Meta` class.

PR #61, opened 2023-06-12
Add possibility to vote on offer

### LINKED ISSUE #38: Vote on offer applications
### Motivation
Core functionality of the tool.

### User Story
As a board member I want to see pending offer applications so that I can vote to admit high quality applications.

### Proposed Solution
- Each user has exactly one vote.
- The vote can be changed until the offer has been finally rejected or accepted. The currently selected option can be highlighted by a check mark.
- An offer will be accepted/rejected when the quorum of voting users is reached.
- The threshold for accepting an offer should be 50% of users that are allowed to vote. As soon as 50,1% accept, the offer should be accepted.
- The threshold for accepted and declined offers should be configurable.

### Screenshots
image

ztefanie:
### Short description
Added site to see and add own votes on applications

### Proposed changes
- added view and template
- added post request handling when new offer is created
- added tests
- added livereload package for hot reload while developing
- added pre-commit package to pinned-dev dependencies, as it was missing
- added code_style.sh 

### Resolved issues

Fixes: #38


integreat_compass/cms/templates/interactions/vote_form.html
```diff
@@ -0,0 +1,67 @@
...
+        {% include "includes/dashboard_navbar.html" %}
+        {% for offer_version in pending_offer_versions %}
+            <form action="{% url 'cms:protected:add_vote' offer_version.id %}"
+                  method="post">
+                {% csrf_token %}
+                <div class="flex flex-row border-2 border-gray-200 rounded-lg shadow-md shadow-inner p-6 mb-8">
+                    <div class="basis-2/3 pr-8">
+                        <p class="text-xl font-bold">{{ offer_version.title }}</p>
+                        <p>{{ offer_version.offer.organization.org_name }}</p>
+                        <p class="text-lg text-gray-700 text-bold pt-4">{% translate "short description" %}</p>
```
JoeyStk:
```suggestion
                        <p class="text-lg text-gray-700 text-bold pt-4">{% translate "Short description" %}</p>
```
Since it's the beginning of the sentence, it should be capatalized

integreat_compass/cms/templates/interactions/vote_form.html
```diff
@@ -0,0 +1,67 @@
...
+                                <p>{% translate "declines" %}</p>
+                                <p class="text-red-600">{{ offer_version.declines }}/{{ number_of_board_members }}</p>
+                            </div>
+                            <div>
+                                <p>{% translate "pending" %}</p>
+                                <p>{{ offer_version.pending }}/{{ number_of_board_members }}</p>
+                            </div>
+                        </div>
+                        <div class="grid grid-cols-3 gap-8 text-white ">
+                            <button class="bg-blue-500 hover:bg-blue-700 rounded-full" type="button">{% translate "to homepage" %}</button>
```
JoeyStk:
```suggestion
                            <a href="{{ organization.web_address }}" class="bg-blue-500 hover:bg-blue-700 rounded-full" type="button">{% translate "to homepage" %}</button>
```
Since I understand this button as a link to the website or Social Media of the organization, that just looks like a button, I would change it to a anker-tag. We store the web_address our model Organization. If you decide to follow this suggestion, you would still need to add the organization to the according view

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,70 @@
...
+                lambda offer_version: offer_version.state
+                == offer_version_states.PENDING,
+                offer_versions,
+            )
+        )
+        for version in pending_offer_versions:
+            version.user_vote = version.votes.filter(creator=request.user).first()
+        return pending_offer_versions
+
+    def get(self, request, *args, **kwargs):
```
JoeyStk: I'm not sure what the best practice for documentation is, but I just wanted to point out that we also added documentation to the get-function [in other places](https://github.com/digitalfabrik/integreat-compass/blob/0843e214ba5f386c3df3fa513e63f6ba15aca647/integreat_compass/cms/views/offers/offer_form_view.py#L56). To keep things as consistent as possible I would suggest to also add documentation here

integreat_compass/cms/models/offers/offer_version.py
```diff
@@ -36,6 +36,7 @@ class OfferVersion(AbstractBaseModel):
         verbose_name=_("Language"),
         help_text=_("The language being taught in this offer"),
     )
+    created_at = models.DateTimeField(default=timezone.now)
```
JoeyStk: Please correct me if I'm wrong, but to me this field seems pretty identical to [offer_version_date](https://github.com/digitalfabrik/integreat-compass/blob/0843e214ba5f386c3df3fa513e63f6ba15aca647/integreat_compass/cms/models/offers/offer_version.py#L20). If the fields do the same thing, I would suggest to either rename the field from offer_version_date to create_at if that makes more sense to you/us or remove the field again

integreat_compass/locale/de/LC_MESSAGES/django.po
```diff
@@ -440,6 +452,50 @@ msgstr "Ankommen weitergedacht."
 msgid "Discover Integreat Compass"
 msgstr "Intgreat Kompass entdecken"
 
+#: cms/templates/interactions/vote_form.html
+msgid "short description"
+msgstr "Kurzbeschreibung"
+
+#: cms/templates/interactions/vote_form.html
+msgid "created at"
+msgstr "Erstellt:"
```
JoeyStk:
```suggestion
msgstr "Erstellt am:"
```
To be a bit closer to the English string I would suggest to add "am"

integreat_compass/cms/templates/interactions/vote_form.html
```diff
@@ -0,0 +1,67 @@
...
+{% endblock title %}
+{% block content %}
+    <div>
+        {% include "includes/dashboard_navbar.html" %}
+        {% for offer_version in pending_offer_versions %}
+            <form action="{% url 'cms:protected:add_vote' offer_version.id %}"
+                  method="post">
+                {% csrf_token %}
+                <div class="flex flex-row border-2 border-gray-200 rounded-lg shadow-md shadow-inner p-6 mb-8">
+                    <div class="basis-2/3 pr-8">
```
JoeyStk, line 16: On mobile and tablet sizes, the content overflows the card a little bit. But since in one of the meetings we said, that this is not highly important for the MVP, I guess it's okay?

integreat_compass/cms/templates/includes/dashboard_navbar.html
```diff
@@ -0,0 +1,20 @@
...
+        </a>
+        {% url 'cms:protected:reports' as url %}
+        <a href="{% url 'cms:protected:reports' %}"
+           class="pr-8 hover:text-gray-900 {% if request.path == url %}font-bold text-gray-950{% endif %}">
+            {% translate "Reports" %}
+        </a>
+        {% url 'cms:protected:declined' as url %}
+        <a href="{% url 'cms:protected:declined' %}"
+           class="hover:text-gray-900 {% if request.path == url %}font-bold text-gray-950{% endif %}">
+            {% translate "Declined" %}
```
JoeyStk, line 17: Are these filters already supposed to work or is this part of a different issue? For me the filters didn't work yet. I got shown the same list in the same order three times

JoeyStk (changes requested):
Thank you very much for the PR :blush: 
All in all this seems to work fine. The main task of the issue was fulfilled. However I have a few ideas or questions. Please find the according comments below :blush:

integreat_compass/cms/templates/interactions/vote_form.html
ztefanie: Hmh, I thought this will link to to detail-page of this offer :thinking: 
This would make a lot of sense, as I maybe want to see all details, e.g. costs of an offer before accepting it. The link to the organization website would be in the details-page as well.

integreat_compass/cms/templates/includes/dashboard_navbar.html
ztefanie: Yes they do nothing now

integreat_compass/cms/models/offers/offer_version.py
ztefanie: Good point, fixed it.

integreat_compass/cms/templates/interactions/vote_form.html
ztefanie: fixed it.

integreat_compass/cms/templates/interactions/vote_form.html
JoeyStk: Ah okay, I understand. I misinterpreted the design then. I think then I find "to website" somewhat misleading, maybe "see more details" would be more intuitive?

charludo: > Hmh, I thought this will link to to detail-page of this offer thinking

Are we going to do this as a modal, like when clicking on an offer in the Filter-view? While it would make sense, a modal on a modal isn't great. On the other hand, openingn the details in its own page would require a new (and so-far unplanned) view

integreat_compass/cms/templates/interactions/vote_form.html
```diff
@@ -0,0 +1,67 @@
...
+                    </div>
+                    <div class="basis-1/3 flex flex-col justify-around align-center font-bold">
+                        {% if offer_version.user_vote %}
+                            {% if offer_version.user_vote.approval %}
+                                <p class="text-green-500">{% translate "You have approved this application" %}</p>
+                            {% else %}
+                                <p class="text-red-500">{% translate "You have declined this application" %}</p>
+                            {% endif %}
+                        {% else %}
+                            <p class="text-gray-700">{% translate "Your vote is still pending" %}</p>
```
charludo:
```suggestion
                            <p class="text-gray-700">{% translate "You have not voted on this application yet" %}</p>
```
I think this is clearer

integreat_compass/cms/urls/protected.py
```diff
@@ -25,5 +25,27 @@
                 ),
             ]
         ),
-    )
+    ),
+    path(
+        "dashboard/",
```
charludo: Should the votes route be below the dashboard route? Should be its own route IMO

integreat_compass/cms/views/interactions/vote_actions.py
```diff
@@ -0,0 +1,35 @@
...
+        offer_version=offer_version_id
+    )
+    if vote:
+        vote[0].approval = approval
+        vote[0].save()
+    else:
+        new_vote = Vote(
+            creator=request.user, offer_version_id=offer_version_id, approval=approval
+        )
+        new_vote.save()
```
charludo:
```suggestion
    vote = Vote.objects.filter(
        creator=request.user, offer_version=offer_version_id
    ).first()
    if vote:
        vote.approval = approval
        vote.save()
    else:
        Vote.objects.create(
            creator=request.user, offer_version_id=offer_version_id, approval=approval
        )

```

integreat_compass/cms/views/interactions/vote_actions.py
```diff
@@ -0,0 +1,35 @@
...
+    return redirect("/dashboard/votes")
```
charludo: Could you please also add a success message for the user? Right now it's really hard to tell if the vote was successfully cast. If your vote crosses the threshold, then the application just seems to disappear.

Ideally, I think different messages should be shown depending on:
- was the vote cast for the first time?
- was the vote changed?
- did the vote lead to acceptance/rejection of the offer application?

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,26 @@
...
+    request = RequestFactory().post("", {"approval": "False"})
+    request.user = User.objects.get(pk=1)
+    add_vote(request, 2)
+    assert Vote.objects.filter(creator=1).filter(offer_version=2).count() == 1
+    assert Vote.objects.filter(creator=1).filter(offer_version=2)[0].approval == False
+
+
+@pytest.mark.django_db
+def test_add_vote_if_exists(load_test_data):
+    assert Vote.objects.filter(creator=1).filter(offer_version=1).count() == 1
```
charludo: This fails for me, `AssertionError: assert 0 == 1`

tests/cms/views/interactions/test_vote_form_view.py
```diff
@@ -0,0 +1,26 @@
...
+from integreat_compass.cms.views.interactions import add_vote, VoteFormView
+
+
+@pytest.mark.django_db
+def test_display_only_newest_offer_version(load_test_data):
+    number_of_board_members = User.objects.filter(groups__name="BOARD_MEMBER").count()
+    request = RequestFactory().get("")
+    request.user = User.objects.get(pk=1)
+    assert OfferVersion.objects.all().count() == 3
+    assert number_of_board_members == 4
```
charludo: This fails for me, `AssertionError: assert 5 == 4`

tools/code_style.sh
```diff
@@ -0,0 +1,28 @@
...
+bash "${DEV_TOOL_DIR}/black.sh"
+
+# Run djlint
+bash "${DEV_TOOL_DIR}/djlint.sh"
+
+# Run pylint
+bash "${DEV_TOOL_DIR}/pylint.sh"
+
+# Run eslint
+# bash "${DEV_TOOL_DIR}/eslint.sh"
```
charludo: What's the reason for commenting this out?

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,87 @@
...
+        :type \**kwargs: dict
+
+        :raises ~django.core.exceptions.PermissionDenied: If user does not have the permission to edit the specific page
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        number_of_board_members = User.objects.filter(
+            groups__name="BOARD_MEMBER"
+        ).count()
```
charludo:
```suggestion
        number_of_board_members = max(User.objects.filter(
            groups__name="BOARD_MEMBER"
        ).count(), settings.NEW_OFFER_GREMIUM_SIZE
```
Requires settings import.

charludo (changes requested):
Thanks, this brings us a huge step forward! 🎉 

Besides the comments below, I have a couple of notes:
- declining the offer application "Deutsch für Anfänger - Intensivkurs" duplicates it in the votes view. Not sure why. This only happens once, and the two applications differ in the number of declines (1 vs 0)
- currently, there's no differentiation between new offer applications and changed offers. The shown `number_of_board_members` should depend on this, and be capped at either `settings.NEW_OFFER_GREMIUM_SIZE` or `settings.CHANGED_OFFER_GREMIUM_SIZE`.

integreat_compass/cms/forms/interactions/vote_form.py
```diff
@@ -0,0 +1,17 @@
+from ...models import Vote
+from ..custom_model_form import CustomModelForm
+
+
+class VoteForm(CustomModelForm):
```
timobrembeck: This form is unused, right? Then it can be removed?
Alternatively, we could also use it to handle the creation/update process of the db objects... but this would involve formsets which can become quite complex^^

integreat_compass/cms/templates/interactions/vote_form.html
```diff
@@ -0,0 +1,67 @@
...
+                            <div>
+                                <p>{% translate "approves" %}</p>
+                                <p class="text-green-600">{{ offer_version.approves }}/{{ number_of_board_members }}</p>
+                            </div>
+                            <div>
+                                <p>{% translate "declines" %}</p>
+                                <p class="text-red-600">{{ offer_version.declines }}/{{ number_of_board_members }}</p>
+                            </div>
+                            <div>
+                                <p>{% translate "pending" %}</p>
```
timobrembeck:
```suggestion
                                <p>{% translate "Approves" %}</p>
                                <p class="text-green-600">{{ offer_version.approves }}/{{ number_of_board_members }}</p>
                            </div>
                            <div>
                                <p>{% translate "Declines" %}</p>
                                <p class="text-red-600">{{ offer_version.declines }}/{{ number_of_board_members }}</p>
                            </div>
                            <div>
                                <p>{% translate "Pending" %}</p>
```
I'd capitalize those

integreat_compass/cms/templates/interactions/vote_form.html
```diff
@@ -0,0 +1,67 @@
...
+                                <p>{{ offer_version.pending }}/{{ number_of_board_members }}</p>
+                            </div>
+                        </div>
+                        <div class="grid lg:grid-cols-3 grid-cols-1 xl:gap-8 gap-2 text-white ">
+                            <button class="bg-blue-500 hover:bg-blue-700 rounded-full" type="button">{% translate "to homepage" %}</button>
+                            <button class="bg-green-500 hover:bg-green-700 rounded-full" type="submit" name='approval' value={{ True }}>
+                                {% translate "accept" %}
+                            </button>
+                            <button class="bg-red-500 hover:bg-red-700 rounded-full" type="submit" name='approval' value={{ False }}>
+                                {% translate "decline" %}
```
timobrembeck:
```suggestion
                                {% translate "Accept" %}
                            </button>
                            <button class="bg-red-500 hover:bg-red-700 rounded-full" type="submit" name='approval' value={{ False }}>
                                {% translate "Decline" %}
```
Those as well

integreat_compass/cms/views/interactions/vote_actions.py
timobrembeck:
```suggestion
    vote, created = Vote.objects.update_or_create(
        creator=request.user,
        offer_version_id=offer_version_id,
        defaults={"approval": approval},
    )
```
Why not simply like this?

integreat_compass/cms/views/interactions/vote_actions.py
```diff
@@ -0,0 +1,35 @@
+from django.shortcuts import redirect
+from django.views.decorators.http import require_POST
+
+from ...decorators import permission_required
+from ...models import Vote
+
+
+@require_POST
+@permission_required("vote_on_offer")
+def add_vote(request, offer_version_id):
```
timobrembeck: I think this function-based view could easily moved to the `post()`-method of the class-based form view? I think it's a bit clearer when form rendering and form submission are part of the same view class.

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,87 @@
...
+        :return: list of offer versions with all details
+        """
+        latest_version_of_offer = OfferVersion.objects.order_by(
+            "offer", "-created_at"
+        ).distinct("offer")
+        offer_versions = (
+            OfferVersion.objects.prefetch_related("votes")
+            .select_related("offer")
+            .select_related("offer__organization")
+            .prefetch_related("offer__tags")
```
timobrembeck:
```suggestion
            OfferVersion.objects.select_related("offer__organization")
            .prefetch_related("offer__tags", "votes")
```
This can be shortened a bit

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,87 @@
...
+            .annotate(approves=Count(Case(When(votes__approval=True, then=1))))
+            .annotate(declines=Count(Case(When(votes__approval=False, then=1))))
+            .annotate(pending=number_of_board_members - Count("votes"))
+            .annotate(user_voted_on=Case(When(votes__creator=request.user, then=True)))
```
timobrembeck:
```suggestion
            .annotate(
                approves=Count("votes", filter=Q(votes__approval=True)),
                declines=Count("votes", filter=Q(votes__approval=False)),
                pending=number_of_board_members - Count("votes"),
                user_voted_on=Count("votes", filter=Q(votes__creator=request.user)),
            )
```
The [docs](https://docs.djangoproject.com/en/4.2/topics/db/aggregation/#filtering-on-annotations) suggest a slightly different syntax, but it doesn't make a huge difference.

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,87 @@
...
+                lambda offer_version: offer_version.state
+                == offer_version_states.PENDING,
+                offer_versions,
+            )
+        )
+        for version in pending_offer_versions:
+            version.user_vote = version.votes.filter(creator=request.user).first()
+        return pending_offer_versions
+
+    def get(self, request, *args, **kwargs):
```
timobrembeck: If you don't perform any actions in the `get()` method besides adding additional template context, it's a bit more clean to just overwrite `get_context_data()` and leave the rendering process up to the base class.

integreat_compass/cms/templates/interactions/vote_form.html
```diff
@@ -0,0 +1,67 @@
+{% extends "_base.html" %}
+{% load i18n %}
+{% load static %}
+{% load widget_tweaks %}
+{% load render_bundle from webpack_loader %}
+{% block title %}
+    "todo"
```
timobrembeck:
```suggestion
    {% translate "Applications" %}
```

integreat_compass/core/settings.py
```diff
@@ -40,6 +40,7 @@
     "django.contrib.contenttypes",
     "django.contrib.sessions",
     "django.contrib.messages",
+    "livereload",
```
timobrembeck: At the moment, this integration is not working yet, so I suggest to either remove it or fix it.

livereload

Also, I'd suggest to add it to the installed apps only when the corresponding package is installed, otherwise this will raise an error in production mode where the dev dependencies are not installed.

pyproject.toml
```diff
@@ -81,6 +81,7 @@ dev-pinned = [
     "coverage==7.2.5",
     "cssbeautifier==1.14.7",
     "dill==0.3.6",
+    "django-livereload-server",
```
timobrembeck: Only adding it to `dev-pinned` without adding it to `dev` will remove it on the next package update.
The idea is that the `dev-pinned` extra is never updated manually, but just via the update script.

integreat_compass/cms/templates/interactions/vote_form.html
ztefanie: Yes I thought we would open (reuse) the modal. What would be the other modal? Don't see why this leads to a modal on a modal situation.

integreat_compass/cms/urls/protected.py
ztefanie: Bildschirmfoto zu 2023-06-22 09-38-57
According to designs it is below dashboard

integreat_compass/cms/views/interactions/vote_actions.py
ztefanie: Nice, did not know about the update_or_create function :+1:

integreat_compass/cms/views/interactions/vote_actions.py
ztefanie: I added a simple success message, if you think more is needed, i think this should be done in a separate ticket and be discussed what is needed / favored.

integreat_compass/cms/forms/interactions/vote_form.py
ztefanie: Deleted it.

integreat_compass/cms/templates/interactions/vote_form.html
charludo: Oh sorry, nevermind, I thought this was openeed through the rejection-comment-modal 🤦🏼‍♀️

integreat_compass/cms/urls/protected.py
charludo: image
Seeing as on the actual dashboard, that link is not highlighted, I think this falls under those "design inconsistencies we should probably ignore"  things 😅

integreat_compass/cms/views/interactions/vote_actions.py
charludo: No, I think that's perfectly fine! Just something of a feedback for the user.

integreat_compass/cms/templates/_base.html
```diff
@@ -1,3 +1,8 @@
+<script>
+    setTimeout(function() {
+        document.getElementById('messages').setAttribute('style', 'display:none !important');
+    }, 3000);
+</script>
```
charludo: If we do want this (personally I don't really see the need), please move it into an appropriate file.

integreat_compass/cms/urls/protected.py
ztefanie: Not sure what I should do here. Where should a user click on in the menu to access the "voting" screens?

integreat_compass/cms/templates/_base.html
ztefanie: I think this is the appropriate file, where do you think this should be?

If we remove it, we will have multiple success messages displayed at once, when a user approves multiple offers.

timobrembeck: Probably @charludo means to put the JS code into a separate `.ts` file and include as part of the webpack bundling process instead of putting it into the html template.

timobrembeck: Also, you could add the class `hidden` instead of setting the style attribute.

integreat_compass/cms/urls/protected.py
charludo: Should be its own route IMO, for example "/applications" or "/approvals" or similar. There's no link on the Dashboard that will take you to the voting screen, and it also wouldn't really fiit from a navigation logic perspective

integreat_compass/cms/templates/_base.html
charludo: Yes, sorry, I meant moving it to its own file like Timo said.

> If we remove it, we will have multiple success messages displayed at once, when a user approves multiple offers.

That still happens if they do so in quick succession. There's other parts around the project as well where multiple error/success messages are shown. Maybe you could make the messages unique (so they do not look like accidental multiples) by changing them to something like `Your vote to approve "TItle of the voted offer" has been saved.`?

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,104 @@
...
+            creator=self.user,
+            offer_version_id=offer_version_id,
+            defaults={"approval": approval},
+        )
+        messages.add_message(
+            self,
+            messages.SUCCESS,
+            _('Your vote on "{}" has successfully been submitted.').format(
+                OfferVersion.objects.get(id=offer_version_id).title
+            ),
```
charludo, line 107: Maybe also indicate the type of vote? (Not strictly necessary though, I'm also fine with this version.)

integreat_compass/cms/urls/protected.py
```diff
@@ -25,5 +25,27 @@
...
+                                name="add_vote",
+                            ),
+                        ]
+                    ),
+                ),
+                path("reports/", interactions.VoteFormView.as_view(), name="reports"),
+                path("declined/", interactions.VoteFormView.as_view(), name="declined"),
+            ]
+        ),
+    ),
```
charludo: Thanks, I think this makes a lot more sense. We could even move the `/applications/votes` view to just `/applications` (since there's nothing there). (And then `applications/votes/add_vote` to `applications/add_vote`, `/applications/reports` to `/reports` and so on.)

Just like the other comment though, not strictly necessary, this would work too.

charludo (changes requested):
Thanks! This works great now! Besides the two (optional) comments below, one issue I just noticed:

The "Applications" tab is shown to all logged in users, not just board members (and staff). While attempting to vote does throw a 403, we should probably also prevent users without voting permissions from even seeing the applications page (and hide the link in the menu)

integreat_compass/cms/urls/protected.py
```diff
@@ -25,5 +25,27 @@
...
+                path(
+                    "votes/",
+                    include(
+                        [
+                            path("", interactions.VoteFormView.as_view(), name="votes"),
+                            path(
+                                "<int:offer_version_id>/add_vote/",
+                                interactions.VoteFormView.add_vote,
+                                name="add_vote",
+                            ),
```
timobrembeck:
```suggestion
```
This was not exactly what I meant by https://github.com/digitalfabrik/integreat-compass/pull/61#discussion_r1229621556, since now you're mixing class-based and function-based views, which is a bit non-standard. I think it would make sense to rename `add_vote()` to `post()` and use the same route for rendering the vote view with `get()` and submit it with `post()` (and pass the offer id either via the post data or as get-parameter).

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,104 @@
...
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        context = super().get_context_data(**kwargs)
+        pending_offer_versions: list = self.get_pending_offer_versions(self.request)
+        context.update({"pending_offer_versions": pending_offer_versions})
+        return context
+
+    @require_POST
+    @permission_required("vote_on_offer")
+    def add_vote(self, offer_version_id):
```
timobrembeck:
```suggestion
    def post(self, request, *args, **kwargs):
```

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,104 @@
...
+from django.shortcuts import redirect
+from django.utils.translation import gettext as _
+from django.views.decorators.http import require_POST
+from django.views.generic import TemplateView
+
+from ...constants import offer_version_states
+from ...decorators import permission_required
+from ...models import OfferVersion, Vote
+
+
```
timobrembeck, line 13:
```suggestion

@method_decorator(permission_required("cms.add_vote"), name="dispatch")
```
As mentioned in #56, I think it makes sense to use the default Django permissions `add_*` etc instead of custom `vote_on_offer` if they essentially mean the same.
Also, we probably can enforce these permissions for the get-view as well, since it does not make sense to render the voting view if the user does not have the permission to cast a vote?

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,104 @@
+from django.contrib import messages
+from django.db.models import Case, Count, When
+from django.shortcuts import redirect
+from django.utils.translation import gettext as _
```
timobrembeck, line 6:
```suggestion
from django.utils.decorators import method_decorator
from django.utils.translation import gettext as _
```

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        Get all offer versions that need a vote from currently logged-in user
+
+        :param request: Object representing the user call
+        :type request: ~django.http.HttpRequest
+
+        :return: list of offer versions with all details
+        """
+        latest_version_of_offer = OfferVersion.objects.order_by(
+            "offer", "-created_at"
```
timobrembeck: Maybe it makes sense to use the default ordering `["offer", "-created_at"]` in the meta class of the offer version model? I guess the `distinct("offer")` will probably be needed quite often?

integreat_compass/cms/templates/interactions/vote_form.html
```diff
@@ -0,0 +1,67 @@
...
+{% block title %}
+    {% translate "Applications" %}
+{% endblock title %}
+{% block content %}
+    <div>
+        {% include "includes/dashboard_navbar.html" %}
+        {% for offer_version in pending_offer_versions %}
+            <form action="{% url 'cms:protected:add_vote' offer_version.id %}"
+                  method="post">
+                {% csrf_token %}
```
timobrembeck, line 14:
```suggestion
                {% csrf_token %}
                <input type="hidden" name="offer_version_id" value="{{ offer_version.id }}">
```

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,104 @@
...
+        :param self: Object representing the user call
+        :type self: ~django.http.HttpRequest
+
+        :param offer_version_id: internal id of the offer_version to be voted on
+        :type offer_version_id: int
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        approval = self.POST.get("approval")
```
timobrembeck:
```suggestion
        approval = request.POST.get("approval")
```

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,104 @@
...
+        Vote.objects.update_or_create(
+            creator=self.user,
+            offer_version_id=offer_version_id,
```
timobrembeck, line 100:
```suggestion
            offer_version_id=request.POST.get("offer_version_id"),
```

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,104 @@
...
+
+        :param offer_version_id: internal id of the offer_version to be voted on
+        :type offer_version_id: int
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        approval = self.POST.get("approval")
+        Vote.objects.update_or_create(
+            creator=self.user,
```
timobrembeck:
```suggestion
            creator=request.user,
```

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,104 @@
...
+            offer_version_id=offer_version_id,
+            defaults={"approval": approval},
+        )
+        messages.add_message(
+            self,
+            messages.SUCCESS,
```
timobrembeck:
```suggestion
        messages.success(
            request,
```

pyproject.toml
```diff
@@ -115,6 +115,7 @@ dev-pinned = [
     "platformdirs==3.5.1",
     "pluggy==1.0.0",
     "pprintpp==0.4.0",
+    "pre-commit==3.3.2",
```
timobrembeck: Adding something to `dev-pinned` manually is not the intended workflow, since this will be lost the next time `./tools/update_dependencies.sh` is executed. Instead, please add the unpinned dependency to the `dev` extra and run the `./tools/update_dependencies.sh` afterwards to update the pinned versions.

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,34 @@
...
+    request.user = User.objects.get(pk=1)
+    setattr(request, "session", "session")
+    messages = FallbackStorage(request)
+    setattr(request, "_messages", messages)
+    return request
+
+
+@pytest.mark.django_db
+def test_add_vote_if_not_exists(load_test_data):
+    assert Vote.objects.filter(creator=1).filter(offer_version=2).count() == 0
```
timobrembeck:
```suggestion
    assert not Vote.objects.filter(creator=1, offer_version=2).exists()
```

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,34 @@
...
+    request = get_request()
+    add_vote(request, 2)
+    assert Vote.objects.filter(creator=1).filter(offer_version=2).count() == 1
+    assert not Vote.objects.filter(creator=1).filter(offer_version=2)[0].approval
```
timobrembeck:
```suggestion
    assert not Vote.objects.get(creator=1, offer_version=2).approval
```
Use `get()` if you expect exactly one matching object and want an error to be thrown if none (or multiple) exist.

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@pytest.mark.django_db
+def test_add_vote_if_exists(load_test_data):
+    assert Vote.objects.filter(creator=1).filter(offer_version=3).count() == 1
+    assert Vote.objects.filter(creator=1).filter(offer_version=3)[0].approval
```
timobrembeck:
```suggestion
    assert Vote.objects.get(creator=1, offer_version=3).approval
```

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,34 @@
...
+    request = get_request()
+    add_vote(request, 3)
+    assert Vote.objects.filter(creator=1).filter(offer_version=3).count() == 1
+    assert not Vote.objects.filter(creator=1).filter(offer_version=3)[0].approval
```
timobrembeck:
```suggestion
    assert not Vote.objects.get(creator=1, offer_version=3).approval
```

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,34 @@
...
+from integreat_compass.cms.views.interactions import add_vote
+
+
+def get_request():
+    request = RequestFactory().post("", {"approval": "False"})
+    request.user = User.objects.get(pk=1)
+    setattr(request, "session", "session")
+    messages = FallbackStorage(request)
+    setattr(request, "_messages", messages)
+    return request
```
timobrembeck: That's probably more of a philosophical discussion, but in my opinion, mocking request objects is not sufficient for testing, because it does not realistically simulate how the user is using the system. Specific settings and middlewares could have side effects which are not covered by these tests, also it does not check whether the URL resolver is configured correctly, it ignores all potential errors that might occur during template rendering, etc...

Testing the specific methods of view classes does provide the additional value that it helps debugging a problem if a test fails (because of the limitations mentioned above, we can e.g. narrow down if it's a bug of the view function itself, or something else), but in my opinion these type of tests help most if there are existing view tests that provide a more realistic scenario.

So in the future, I think it would make sense to start with writing tests using the [test client](https://docs.djangoproject.com/en/4.2/topics/testing/tools/#the-test-client) and verifying whether making actual requests has the effect we expect. And only if time is left, we can improve these tests and write more fine-graded tests which help to identify the actual problem when a tests fails.

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+from django.contrib.messages.storage.fallback import FallbackStorage
+from django.test import RequestFactory, TestCase
+
+from integreat_compass.cms.models import User, Vote
+from integreat_compass.cms.views.interactions import add_vote
```
timobrembeck: This needs to be adapted to the new location

tests/cms/views/interactions/test_vote_form_view.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from django.test import RequestFactory, TestCase
+
+from integreat_compass.cms.models import OfferVersion, User, Vote
+from integreat_compass.cms.views.interactions import add_vote, VoteFormView
```
timobrembeck: This as well

tests/cms/views/interactions/test_vote_form_view.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@pytest.mark.django_db
+def test_display_only_newest_offer_version(load_test_data):
+    request = RequestFactory().get("")
+    request.user = User.objects.get(pk=1)
+    assert OfferVersion.objects.all().count() == 3
+    pending_offer_versions = VoteFormView.get_pending_offer_versions(request, 4)
```
timobrembeck:
```suggestion
    pending_offer_versions = VoteFormView.get_pending_offer_versions(request)
```

integreat_compass/cms/templates/interactions/vote_form.html
timobrembeck:
```suggestion
                            <button class="bg-blue-500 hover:bg-blue-700 rounded-full" type="button">{% translate "Details" %}</button>
```
I agree with @JoeyStk that "to homepage" is not a good label if it does not lead to the homepage.

timobrembeck (changes requested):
Thanks for your changes!

integreat_compass/cms/views/interactions/vote_form_view.py
ztefanie: Keeped it, as it is now used.

timobrembeck: I didn't want to remove it, but only add the decorator import as well.

integreat_compass/cms/views/interactions/vote_form_view.py
ztefanie: addded it to the meta.

integreat_compass/cms/templates/interactions/vote_form.html
ztefanie: If i remove this, I get the error "Reason given for failure:    CSRF token missing."
So i kept it.

timobrembeck: Look closely at the suggestion, I did not want to remove the line, but add another one. Github just renders the diff a bit weird with suggestions. But you don't have to copy this manually, if you directly commit the suggestion, it should work that the new line is inserted while the old line stays unchanged.

tests/cms/views/interactions/test_vote_action.py
ztefanie: I thought writing `count() == 1` is more clear ;)

timobrembeck: Hmm, I disagree. You use `count() == 1` when you want to ensure that the query returns exactly one element without caring about what the element actually is. In this case, this is not true, since we also are interested in the attributes of the returned object (in this case the `approved` field). If you want to check existence and approval in two separate statements, I would prefer the following syntax:
```
vote = Vote.objects.get(creator=1, offer_version=2)
assert not vote.approval
```

integreat_compass/cms/templates/_base.html
```diff
@@ -11,7 +12,11 @@
                      alt="{% translate "Logo" %}"
                      class="h-24 w-48">
             </a>
-            <a href="#" class="ml-6 mr-4 hover:text-gray-900">{% translate "Dashboard" %}</a>
+            <a href="" class="ml-6 mr-4 hover:text-gray-900">{% translate "Dashboard" %}</a>
+            {% if user|has_group:"BOARD_MEMBER" %}
```
timobrembeck:
```suggestion
            {% if perms.cms.add_vote %}
```
Groups are just an abstraction to make it easier to assign the same set of permissions to multiple users, however in the templates & views we want to check for the actual permissions instead of group memberships, so even if we change the permissions of a group in the future, the functionality still stays consistent.

integreat_compass/cms/models/offers/offer.py
```diff
@@ -40,7 +40,7 @@ def public_version(self):
         :return: OfferVersion or ``None``
         :rtype: ~integreat_compass.cms.models.offers.offer_version.OfferVersion
         """
-        versions = self.versions.order_by("offer_version_date")
+        versions = self.versions.order_by("created_at")
         for version in versions:
```
```suggestion
        for version in self.versions.all():
```
Ordering is no longer required, since this already happens in the meta class of the model (which automatically applies the ordering to all querysets, including related managers).

integreat_compass/cms/templates/interactions/vote_form.html
```diff
@@ -0,0 +1,67 @@
...
+                                <p>{{ offer_version.pending }}/{{ offer_version.number_of_votes_needed }}</p>
+                            </div>
+                        </div>
+                        <input type="hidden" name="offer_version_id" value={{ offer_version.id }}>
+                        <div class="grid lg:grid-cols-3 grid-cols-1 xl:gap-8 gap-2 text-white ">
+                            <button class="bg-blue-500 hover:bg-blue-700 rounded-full" type="button">{% translate "Details" %}</button>
+                            <button class="bg-green-500 hover:bg-green-700 rounded-full" type="submit" name='approval' value={{ True }}>
+                                {% translate "Accept" %}
+                            </button>
+                            <button class="bg-red-500 hover:bg-red-700 rounded-full" type="submit" name='approval' value={{ False }}>
```
timobrembeck:
```suggestion
                        <input type="hidden" name="offer_version_id" value="{{ offer_version.id }}">
                        <div class="grid lg:grid-cols-3 grid-cols-1 xl:gap-8 gap-2 text-white">
                            <button class="bg-blue-500 hover:bg-blue-700 rounded-full" type="button">{% translate "Details" %}</button>
                            <button class="bg-green-500 hover:bg-green-700 rounded-full" type="submit" name="approval" value="{{ True }}">
                                {% translate "Accept" %}
                            </button>
                            <button class="bg-red-500 hover:bg-red-700 rounded-full" type="submit" name="approval" value="{{ False }}">
```
I'd prefer consistent quoting of html attributes (also it reduces the potential of errors when the context variables contain spaces)

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,34 @@
...
+    request.user = User.objects.get(pk=1)
+    setattr(request, "session", "session")
+    messages = FallbackStorage(request)
+    setattr(request, "_messages", messages)
+    return request
+
+
+@pytest.mark.django_db
+def test_add_vote_if_not_exists(load_test_data):
+    assert not Vote.objects.filter(creator=1).filter(offer_version=2).exists()
```
timobrembeck:
```suggestion
    assert not Vote.objects.filter(creator=1, offer_version=2).exists()
```

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,34 @@
...
+    request = get_request(2)
+    VoteFormView.as_view()(request)
+    assert Vote.objects.filter(creator=1).filter(offer_version=2).get()
+    assert not Vote.objects.filter(creator=1).filter(offer_version=2)[0].approval
```
timobrembeck:
```suggestion
    vote = Vote.objects.get(creator=1, offer_version=2)
    assert not vote.approval
```
- `filter().filter().get()` does not really make sense - just use `get()` in the first place.
- `filter()[0]` does not make sense - either use `get()` if you know exactly one element exists or `filter().first()` if you're unsure how many objects might exist with these filters.

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,110 @@
...
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        context = super().get_context_data(**kwargs)
+        pending_offer_versions: list = self.get_pending_offer_versions(self.request)
+        context.update({"pending_offer_versions": pending_offer_versions})
+        return context
+
+    @transaction.atomic
+    def post(self, request, *args, **kwargs):
+        """
```
timobrembeck:
```suggestion
        r"""
```

integreat_compass/locale/de/LC_MESSAGES/django.po
```diff
@@ -460,6 +473,51 @@ msgstr "Ankommen weitergedacht."
...
+#: cms/templates/interactions/vote_form.html
+msgid "Declines"
+msgstr "Absagen"
+
+#: cms/templates/interactions/vote_form.html
+msgid "Pending"
+msgstr "Offen"
+
+#: cms/templates/interactions/vote_form.html
+#| msgid "Offer details"
```
timobrembeck:
```suggestion
```

timobrembeck (changes requested):

integreat_compass/cms/migrations/0002_alter_offerversion_options.py
timobrembeck, line 1: Since you're modifying the initial migration anyways, I think we can also move these changes into 0001 as well...

integreat_compass/cms/migrations/0001_initial.py
```diff
@@ -385,7 +382,7 @@ class Migration(migrations.Migration):
             options={
                 "verbose_name": "offer version",
                 "verbose_name_plural": "offer versions",
-                "ordering": ["-offer_version_date"],
+                "ordering": ["-created_at"],
```
timobrembeck:
```suggestion
                "ordering": ["offer", "-created_at"],
```

integreat_compass/cms/templates/_base.html
```diff
@@ -1,6 +1,7 @@
 {% extends "_raw.html" %}
 {% load i18n %}
 {% load static %}
+{% load user_filters %}
```
timobrembeck:
```suggestion
```
These are no longer required

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,110 @@
...
+            defaults={"approval": approval},
+        )
+        messages.add_message(
+            request,
+            messages.SUCCESS,
+            _('Your vote on "{}" has successfully been submitted.').format(
+                OfferVersion.objects.get(id=offer_version_id).title
+            ),
+        )
+        return redirect("/applications/")
```
timobrembeck:
```suggestion
        return redirect("cms:protected:votes")
```
The only place where you use hardcoded paths is the url config. In all other places, names url patterns are preferred.

tests/cms/views/interactions/test_vote_action.py
```diff
@@ -0,0 +1,34 @@
...
+    return request
+
+
+@pytest.mark.django_db
+def test_add_vote_if_not_exists(load_test_data):
+    assert not Vote.objects.filter(creator=1, offer_version=2).exists()
+    request = get_request(2)
+    VoteFormView.as_view()(request)
+    assert Vote.objects.get(creator=1, offer_version=2)
+    assert not Vote.objects.filter(creator=1, offer_version=2).first().approval
```
timobrembeck:
```suggestion
    assert not Vote.objects.get(creator=1, offer_version=2).approval
```

Both of these lines do not really make sense, because of similar reasons in my previous comments.
- `assert Vote.objects.get()` does never run the `assert` statement, since there are only two possible outcomes of `get()`: An object (which always evaluates to `True`), or an exception (which lets the test fail before the `assert` even runs).
- Accessing attributes after `filter().first()` does not make sense since the only reason why you would prefer `filter().first()` over `get()` is when you want to tolerate `None` as results In this case, if `None` would be returned, this would throw an exception `AttributeError: 'NoneType' object has no attribute 'approval'`, which is a far worse error message than the `ObjectDoesNotExist` which is thrown by `get()` if no such object exists.

I'm not quite sure what part of my suggestion you don't like, could you explain a bit why you would prefer your solution? Am I missing something here?

timobrembeck (changes requested):
Thanks a lot, just a few small details left! :+1:

And could you resolve the conflict in the `package-lock.json`?

tests/cms/views/interactions/test_vote_action.py
ztefanie: I thought I just did what you suggested, I just want to "assert" that the value of approval has been updated.

timobrembeck (commented):
Thanks! :pray:

Could you resolve the conflicts yet again?

integreat_compass/cms/views/interactions/vote_form_view.py
```diff
@@ -0,0 +1,110 @@
...
+        offer_version_id = request.POST.get("offer_version_id")
+        approval = request.POST.get("approval")
+        Vote.objects.update_or_create(
+            creator=request.user,
+            offer_version_id=offer_version_id,
+            defaults={"approval": approval},
+        )
+        messages.add_message(
+            request,
+            messages.SUCCESS,
```
timobrembeck:
```suggestion
        messages.success(
            request,
```

integreat_compass/cms/models/offers/offer_version.py
```diff
@@ -113,4 +127,5 @@ class Meta:
         verbose_name = _("offer version")
         verbose_name_plural = _("offer versions")
         default_related_name = "offer_versions"
-        ordering = ["-offer_version_date"]
+        ordering = ["offer", "-created_at"]
+        default_permissions = ("change", "delete", "view")
```
```suggestion
        default_permissions = ()
```
Do we even need these permissions at all? Or could we just argue, that we use the offer permissions for offer versions as well?

timobrembeck:
Sorry, I think we need one rebase again to fix the tests :see_no_evil:

timobrembeck (approved):
Thanks a lot! :pray:

charludo (approved):

JoeyStk (approved):
:rocket:

PR #63, opened 2023-06-12
Add map to offer application

### LINKED ISSUE #50: Create Offer Location
### Motivation
Offers should be findable by location.

### User Story
As a organization representative I want enter a location for an offer ("Maßnahme") so that the offer can be found by job center employees by distance.

### Proposed Solution
- The user can drag and drop the marker on a map. We only store longitude/latitude.
- The user can enter a description that is not validated (saved "as is").
- For the MVP we do not use any API to resolve location to coordinates or vice versa.
- The location is only relevant for non-online offers. The location is mandatory for presence offers. However, the user is not required to enter a specific address. It would be okay to just use the coordinates of a city or county.
- The location map should use our maps.tuerantuer.org server and not Google Maps. We could use OpenStreetMap Nomatim for resolving addresses.

### Additional Context
The general offer form is covered by #37 

### Screenshots
image

charludo:
### Short description

Add map to offer application

### Proposed changes

- add the map
- address field can be used to search, and will in turn be populated if user clicks/drags marker on map

### Side effects

- I noticed we had prettier's width set to only 80 chars. This breaks up a lot of TS lines and makesthe code much less readable, so I changed it to 120 (same as for Integreat CMS). Unfortunately this somewhat bloats the diff of this PR.
- some values are hardcoded (center of map, Nominato URL). I guess we could also make them configurable, I thought it wasn't really worth the extra effort right now.

### Resolved issues

Fixes: #50

### DIFF
--- a/integreat_compass/static/src/js/map.ts
+++ b/integreat_compass/static/src/js/map.ts
@@ -1,0 +1,119 @@
+import { FullscreenControl, LngLat, LngLatLike, Map, MapOptions, MapMouseEvent, Marker } from "maplibre-gl";
+import { getAddressFromCoordinates, getCoordinatesFromAddress } from "./nominatim";
+import "maplibre-gl/dist/maplibre-gl.css";
+
+const getCoordinateInputFields = () => {
+    const latitudeInput = document.getElementById("id_location-lat") as HTMLInputElement;
+    const longitudeInput = document.getElementById("id_location-long") as HTMLInputElement;
+    return [latitudeInput, longitudeInput];
+};
+
+const getCoordinates = () => {
+    const [latitudeInput, longitudeInput] = getCoordinateInputFields();
+    if (latitudeInput.value && longitudeInput.value) {
+        return new LngLat(longitudeInput.valueAsNumber, latitudeInput.valueAsNumber);
+    }
+    return null;
+};
+
+export const updateField = (fieldName: string, value: number) => {
+    const field = document.getElementById(`id_location-${fieldName}`) as HTMLInputElement;
+    if (value && field.value !== value.toString()) {
+        field.value = value.toString();
+    }
+};
+
+const updateCoordinates = (marker: Marker) => {
+    const lngLat = marker.getLngLat();
+    updateField("long", lngLat.lng);
+    updateField("lat", lngLat.lat);
+};
+
+window.addEventListener("load", () => {
+    /* eslint-disable-next-line no-magic-numbers */
+    const centerOfGermany = [10.3873, 51.215];
+    const centeredZoom = 16;
+    const minAddressLength = 3;
+
+    const coordinates = getCoordinates();
+    const container = document.getElementById("map");
+    const searchBar = document.getElementById("id_location-address") as HTMLInputElement;
+    const modeSwitch = document.querySelector(".mode_type") as HTMLElement;
+    const onlineCheckbox = document.getElementById("id_offer-mode_type_0") as HTMLInputElement;
+
+    if (!container || !searchBar || !onlineCheckbox) {
+        return;
+    }
+
+    const options = {
+        container,
+        style: "https://maps.tuerantuer.org/styles/integreat/style.json",
+    } as MapOptions;
+
+    if (coordinates) {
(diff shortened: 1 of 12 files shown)


JoeyStk (approved):
Thank you very much. Works as expected!

I just wanted to raise the question how fussy we should be about disabling linters. Is it okay to make a few exceptions? I don't have a strong opinion on that, but just want to know what you think about it

package.json
```diff
@@ -37,5 +37,8 @@
     "webpack": "^5.76.0",
     "webpack-bundle-tracker": "^1.8.1",
     "webpack-cli": "^5.0.1"
+  },
+  "dependencies": {
+    "maplibre-gl": "^3.0.1"
```
timobrembeck: Since we don't use npm in production, I think it makes sense to add this to the dev dependencies instead...

webpack.config.js
```diff
@@ -87,5 +85,12 @@ module.exports = {
 
   plugins: [
     new BundleTracker({ filename: "./integreat_compass/webpack-stats.json" }),
+    new MiniCssExtractPlugin({
+      filename: function (config) {
+        if (config.chunk.name == "pdf") return "[name].css";
```
timobrembeck: as long as we don't have special css for pdfs, I guess we don't need this exception?

integreat_compass/static/src/js/map.ts
```diff
@@ -0,0 +1,114 @@
...
+        updateCoordinates(marker);
+        updateAddress();
+    });
+    marker.on("dragend", () => {
+        updateCoordinates(marker);
+        updateAddress();
+    });
+
+    searchBar.addEventListener("keypress", (event) => {
+        if (event.key === "Enter") {
```
timobrembeck, line 100: Why only when the Enter key is pressed? To me, this is a bit counter-intuitive, since I expect the enter key to submit the form, and not that it updates the map.
I think e.g. the `focusout` event would be a better fit?

integreat_compass/static/src/js/nominatim.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const getAddressFromCoordinates = async (ll: LngLat): Promise<string> => {
+    const url = `${apiBaseUrl}reverse?format=json&lat=${ll.lat}&lon=${ll.lng}`;
+
+    try {
+        const response = await fetch(url);
+        if (response.ok) {
+            const data = await response.json();
+            /* eslint-disable camelcase */
+            const { road, house_number, postcode, city } = data.address;
```
timobrembeck: `city` only works for larger cities, smaller ones are often times referred to as `town` or `village` by nominatim...
but ok, since the `display_name` is more or less usable as fallback, I guess it's ok like this.

timobrembeck (commented):
Thanks, works as expected! :+1:

Just a few minor comments...

> I just wanted to raise the question how fussy we should be about disabling linters. Is it okay to make a few exceptions? I don't have a strong opinion on that, but just want to know what you think about it

@JoeyStk I think it's pretty hard to define consistent standards here, I think it mostly depends on each individual case, often times it is better to fix the issue than to ignore the rule, but with the two exceptions made in this PR I absolutely agree...

integreat_compass/static/src/js/nominatim.ts
charludo: Yeah I noticed that as well. This is not really a well-rounded solution, I just felt that SOMETHING should be put as the address if a location is selected via the map.

I'll have a look at using town and village as "fallbacks" for city.

(Funnily enough, for the town I'm living in, all those fields are enmpty. Just A roadname and postcode.)

integreat_compass/static/src/js/map.ts
charludo: Generally I would agree, but we are essentially presenting the address field as a search field. `focusout` is a good idea, but I'd advocate for using it in addtion, not instead

webpack.config.js
charludo: Whoops.

timobrembeck (approved):
:tada:

JoeyStk (approved):

PR #65, opened 2023-06-14
Create new OfferVersion when details were changed

### LINKED ISSUE #64: Actually create a new OfferVersion instead of editing it
### Motivation

Some changes of offer details require re-approval from board members. We extracted the affected fields into the `OfferVersion` model. However, making changes to `OfferVersion` fields in the `OfferFormView` currently simply updates the existing instance.

### Proposed Solution

In the `OfferFormView` post method, check if the `OfferVersionForm` contains changes. If so, create a new `OfferVersion` instead of  saving the form.

### Alternatives

Alternatively, it might also be possible to simply overwrite the `OfferVersionForm` `save()` method to create and return a new instance, instead of saving the old one. If possible, this might be the cleaner approach.

charludo:
### Short description

Previously we just updated the version, which we specifically did not want to do

### Proposed changes

- check if any offer version fields were changed by the user; if so, create a new instance
- actually get the latest offer version instance in the offer form view

### Side effects

/

### Resolved issues

Fixes: #64

### DIFF
--- a/integreat_compass/cms/views/offers/offer_form_view.py
+++ b/integreat_compass/cms/views/offers/offer_form_view.py
@@ -33,7 +33,7 @@ class OfferFormView(TemplateView):
     def _get_forms(offer_id, **kwargs):
         if offer_id:
             offer = Offer.objects.get(id=offer_id)
-            offer_version = offer.versions.first()
+            offer_version = list(offer.versions.select_related())[-1]
             offer_contact = offer.offer_contact
             offer_location = offer.location
             offer_organization = offer.organization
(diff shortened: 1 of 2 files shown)


integreat_compass/cms/views/offers/offer_form_view.py
```diff
@@ -32,7 +32,7 @@ class OfferFormView(TemplateView):
     def _get_forms(offer_id, **kwargs):
         if offer_id:
             offer = Offer.objects.get(id=offer_id)
-            offer_version = offer.versions.first()
+            offer_version = list(offer.versions.select_related())[-1]
```
charludo, line 36: This is incredibly ugly, but I could not figure out a different way. The ordering won't budge when using `order_by`, and using `reverse` on the selected offers has no effect either.

Any ideas how to improve this?

charludo: Update:  #71 uses the same approach but inside a latest_version property, so probably move the discussion to there.

JoeyStk (approved):
Seems to do the trick. Thank you!

PR #66, opened 2023-06-14
Fix over-precise coordinates preventing form submission

charludo:
### Short description

Small fix for a bug I did not notice before: Previously, when clicking on the map, the coordinates received from the marker had larger post-comma precision than the coordinate field on the location model allowed for, and the form would not Save. The user would also not be notified, since the lat/long fields are hidden.

### Proposed changes

- increase precision in the `Location` model
- fix the coordinates received from the marker to the same number of decimal places


JoeyStk (approved):
This seems to fix the issue. Thank you!

PR #69, opened 2023-06-16
Offer Application usability tweaks

charludo:
### Short description

This PR addresses some usability issues with the offer application form Martin pointed out.

### Proposed changes

- adjusted the help texts for Contact fields to be clearer (esp. in English)
- pre-select German as the default offer language
- add a scroll-offset to input elements, so when client-side validation fails and the browser jumps to a field, it isn't covered by our menu bar
- auto-fill `http://` in front of the web address if the user did not specify a protocol (client-side validation fails without it) 

### Side effects

/ 

### Resolved issues

Fixes: /


svenseeberg (approved):
Good improvements indeed, thanks!

PR #70, opened 2023-06-16
Flexible Gremium size

### LINKED ISSUE #59: Flexible Gremium size / vote threshold
### Motivation
We don't know how many persons will initially work with the MVP. It is possible that we do not have 6 persons for voting on offer applications.

### User Story
N/A

### Proposed Solution
We may need to make the threshold for accepting offer applications flexible. That means that `NEW_OFFER_GREMIUM_SIZE` can change over time. However, accepted offers should not be removed from the list, just because the number of voting users changed. That means the status if an offer is accepted or not needs to be stored with the offer version at the time of the last relevant vote ("accepted by a majority").

### Alternatives
NEW_OFFER_GREMIUM_SIZE could be time based. But that would be even more complicated.

### Additional Context
We can wait with this until we know how many persons will participate in testing the MVP.

charludo:
### Short description

Ensure that changing gremium size settings will not change the state of already-accepted/rejected offers.

### Proposed changes

- create post-save signal for Vote which determines and sets the new state of the offer version
- change state to be a field of OfferVersion, rather than a property

### Side effects

- if greium members were to vote on an already-accepted/rejected offer after the gremium size was changed, we could run into the same issue as before. However, I think this is a non-issue if we just don't present offers which have been decided upon to the members for further voting. 

### Resolved issues

Fixes: #59


integreat_compass/cms/models/offers/offer_version.py
```diff
@@ -60,6 +60,11 @@ class OfferVersion(AbstractBaseModel):
         verbose_name=_("Language"),
         help_text=_("The language being taught in this offer"),
     )
+    state = models.CharField(
```
```suggestion
    state = models.BooleanField(
```
The data types of the underlying constants are Boolean, so either change the field type, or the type of the constants.

integreat_compass/core/signals/vote_signals.py
```diff
@@ -0,0 +1,44 @@
...
+        return offer_version_states.APPROVED
+
+    if offer_version.votes.filter(approval=False).count() > max_votes // 2:
+        return offer_version_states.REJECTED
+
+    return offer_version_states.PENDING
+
+
+@receiver(post_save, sender=Vote)
+def update_offer_version_state(instance, **kwargs):
```
timobrembeck, line 33: Maybe we need database transactions here to prevent race conditions between simultaneous votes?

timobrembeck (approved):
Thanks! :+1:
Didn't try it out with the different gremium sizes, but the code looks good!

PR #71, opened 2023-06-16
My Offers view

### LINKED ISSUE #46: View my offers
### Motivation
We need a way for users to manage offers.

### User Story
As an organization representative I want to see all my offers and their status.

### Proposed Solution
- A simple list of all applications with status "pending", "accepted", "denied".
- Button to create new offer applications.
- Delete existing offers or applications.
- It should be possible to edit the location and contact info, type of offer. Categories, description, title, organisation name cannot be changed. If that needs to be changed, the user has to create a new application. We can add a sentence to the form that explains this process.

### Screenshots
image

charludo:
### Short description

Add "My offers" view and possibility to delete offers

### Proposed changes

- add view and template 
- I took someliberties with the design, see comments on the issue.

### Side effects

- none I am aware of

### Resolved issues

Fixes: #46


integreat_compass/cms/templates/offers/my_offers_list.html
```diff
@@ -0,0 +1,66 @@
...
+                             src="{{ offer.latest_version.title_image.url }}"
+                             alt="Title image of this offer"
+                             width="300"
+                             height="300">
+                    </div>
+                    <div class="pr-8 py-6">
+                        <p class="text-xl font-bold">{{ offer.latest_version.title }}</p>
+                        <p>{{ offer.organization.org_name }}</p>
+                        <p class="text-gray-600 mt-4">{{ offer.latest_version.description|truncatechars:500 }}</p>
+                        <div class=" my-4">
```
JoeyStk:
```suggestion
                        <div class="my-4">
```

integreat_compass/cms/templates/offers/my_offers_list.html
```diff
@@ -0,0 +1,66 @@
...
+                            {% for tag in offer.tags.all %}
+                                <span class="bg-red-500 text-white rounded-full py-1 px-2 mr-1">{{ tag }}</span>
+                            {% endfor %}
+                        </div>
+                        <p class="text-gray-700">{% translate "created at" %}: {{ offer.latest_version.offer_version_date }}</p>
```
JoeyStk:
```suggestion
                        <p class="text-gray-700">{% translate "Created at:" %} {{ offer.latest_version.offer_version_date }}</p>
```

JoeyStk (dismissed):
Thank you!
All in all this LGTM. 
I found two minor issues, please see my suggestions for them below :)

integreat_compass/cms/views/offers/my_offers_list_view.py
```diff
@@ -0,0 +1,85 @@
...
+        :type \**kwargs: dict
+
+        :raises ~django.core.exceptions.PermissionDenied: If user does not have the permission to edit the specific page
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+
+        offers = Offer.objects.filter(creator=request.user)
+        logger.warning(offers)
```
svenseeberg: Is that actually a warning? :)

svenseeberg (changes requested):
I found one remaining debug output, that should probably be removed.

integreat_compass/cms/views/offers/my_offers_list_view.py
charludo: Whoopsie :D

integreat_compass/cms/models/offers/offer.py
```diff
@@ -46,6 +46,16 @@ def public_version(self):
...
 
+    @cached_property
+    def latest_version(self):
+        """
+        Returns the latest version of an offer.
+
+        :return: OfferVersion
+        :rtype: ~integreat_compass.cms.models.offers.offer_version.OfferVersion
+        """
+        return list(self.versions.select_related())[-1]
```
```suggestion
        return self.versions.select_related().first()
```
Retrieving all versions (including all related models) is a bit overkill just for the latest version. It's more efficient to perform the filtering on the sql layer and just retrieve the one record we're actually interested in.
If you tried this and it didn't work for you, it is because the date was not correctly set on a newer offer version. But if the version creation is fixed with the correct dates, the `first()` will actually return the latest version.

integreat_compass/cms/views/offers/my_offers_list_view.py
```diff
@@ -0,0 +1,84 @@
...
+
+@method_decorator(permission_required("cms.change_offer"), name="dispatch")
+class MyOffersListView(TemplateView):
+    """
+    View showing all of a user's offers
+    """
+
+    template_name = "offers/my_offers_list.html"
+
+    def get(self, request, *args, **kwargs):
```
timobrembeck: If you're only adding context and don't change the rendering process, it makes more sense to override `get_context_data()` instead of `get()`.

integreat_compass/cms/views/offers/my_offers_list_view.py
```diff
@@ -0,0 +1,84 @@
...
+from django.views.generic import TemplateView
+
+from ...decorators import permission_required
+from ...models import Offer
+
+logger = logging.getLogger(__name__)
+
+
+@method_decorator(permission_required("cms.change_offer"), name="dispatch")
+class MyOffersListView(TemplateView):
```
timobrembeck:
```suggestion
class MyOffersListView(ListView):
```
I think it would make sense to let this inherit from Django's inbuilt [`ListView`](https://docs.djangoproject.com/en/4.2/ref/class-based-views/generic-display/#listview) and just override `get_queryset()` to filter the offers according to the current user.

integreat_compass/cms/views/offers/my_offers_list_view.py
```diff
@@ -0,0 +1,84 @@
+import logging
+
+from django.contrib import messages
+from django.db import transaction
+from django.shortcuts import redirect, render
+from django.utils.decorators import method_decorator
+from django.utils.translation import gettext as _
+from django.views.generic import TemplateView
```
timobrembeck:
```suggestion
from django.views.generic.list import ListView
```

integreat_compass/cms/views/offers/my_offers_list_view.py
```diff
@@ -0,0 +1,84 @@
...
+
+        offers = Offer.objects.filter(creator=request.user)
+        return render(
+            request,
+            self.template_name,
+            {**self.get_context_data(**kwargs), "offers": offers},
+        )
+
+    @transaction.atomic
+    def post(self, request, *args, **kwargs):
```
timobrembeck: I think it makes sense to outsource the deletion into a separate view in case you want to allow deletion from other views as well.
You could e.g. create a new view `OfferDeleteView` which inherits from [`DeleteView`](https://docs.djangoproject.com/en/4.2/ref/class-based-views/generic-editing/#deleteview) and then you only have to override `form_valid` to show the success message.

And if you want to make sure a user can only delete their own offers, it would make more sense to also add [`PermissionRequiredMixin`](https://docs.djangoproject.com/en/4.2/topics/auth/default/#the-permissionrequiredmixin-mixin) as second base class and overwrite `get_permission_required()` to make sure to raise a 403 error if a user tries to delete a different offer than their own...

tools/prune_database.sh
```diff
@@ -49,4 +49,6 @@ fi
 
 # Remove media files (because they are no longer usable without the corresponding database entries)
 rm -rfv "${PACKAGE_DIR:?}/media"
+mkdir "${PACKAGE_DIR:?}/media"
+cp "${PACKAGE_DIR:?}/static/images/fallback_title_image.png" "${PACKAGE_DIR?:}/media/fallback_title_image.png"
```
timobrembeck: Does this make sense? I think we should rather adapt the fallback path to reference the image in `/static/` than to move an image into the media directory which is not handled by the normal Django upload storage. This will also cause overhead on all production systems if we have to move this file manually.

integreat_compass/cms/models/offers/offer.py
charludo: Does this work for you? I know it should, but just did not for me.

timobrembeck: Could you head over to http://localhost:8082/admin/cms/offerversion/ and inspect the `offer_version_date` fields of the offer versions? In my case, all versions had the same date. After fixing the date and give the latest version a newer date, it worked.
So when creating a new offer version, we have to not only increment the id, but also set the date to the current date...

timobrembeck: So this seems to be a bug introduced by #65.

charludo: Ah, alright. Thank you for catching this!!

tools/prune_database.sh
charludo: Yes, this is a really ugly workaround.
Unfortunately I can not get the fallback image to be looked for anywhere but the media dir. I re-did the changes I originally thought would work, could you have another look?

(Thanks for all the reviews btw!)

timobrembeck: Ah ok I get it, sorry I wasn't aware that the default value is stored on the image field.
Hmm, what do you think about creating a new property on the offer model and leave the image field in the db empty per default? Something like this:
```
def title_image_url(self):
    return self.title_image.url if self.title_image else settings.DEFAULT_TITLE_IMAGE
```
This would require to replace `title_image.url` with `title_image_url` in the templates and to remove the custom handling in the forms, so not sure whether it's worth it.
I would also be fine with the previous solution of copying the file manually...

charludo: I like the idea, but would prefer keeping the handling in the forms; I'll try to figure out a way to do both.

charludo: Alright, I think I found an acceptable workaround. The field is now nullable, and the templates just have to check if the images exist, otherwise display the static fallback image.

svenseeberg (approved):
Looks good IMHO :) Thanks

charludo:
@JoeyStk is there anything else you want to change?

integreat_compass/cms/forms/offers/offer_version_form.py
```diff
@@ -82,7 +77,7 @@ def clean(self):
         """
         cleaned_data = super().clean()
         if cleaned_data["title_image"] is False:
-            cleaned_data["title_image"] = settings.DEFAULT_TITLE_IMAGE
+            cleaned_data["title_image"] = ""
```
timobrembeck:
```suggestion
            cleaned_data["title_image"] = None
```
Wouldn't this be `None` instead of an empty string?

integreat_compass/cms/templates/offers/my_offers_list.html
```diff
@@ -0,0 +1,67 @@
...
+        <h1 class="mb-8">{% translate "My offers" %}</h1>
+        {% for offer in object_list %}
+            <form method="post"
+                  action="{% url 'cms:protected:delete_offer' offer.pk %}">
+                {% csrf_token %}
+                <input type="hidden" name="pk" value="{{ offer.pk }}">
+                <div class="md:flex flex-row border-2 justify-between border-gray-200 rounded-lg shadow-md shadow-inner mb-8 gap-6 overflow-hidden">
+                    <div class="basis-1/8 shrink-0">
+                        <img class="object-cover h-[300px]"
+                             src="{% if offer.latest_version.title_image and offer.latest_version.title_image.url %}{{ offer.latest_version.title_image.url }}{% else %}{% static fallback_title_image %}{% endif %}"
```
timobrembeck:
```suggestion
                             src="{% if offer.latest_version.title_image %}{{ offer.latest_version.title_image.url }}{% else %}{% static fallback_title_image %}{% endif %}"
```
If `offer.latest_version.title_image` evaluates to `True`, it is guaranteed that `offer.latest_version.title_image.url` does as well.

Also, I think we will probably need this expression multiple times, so why not add a property for this as suggested in https://github.com/digitalfabrik/integreat-compass/pull/71#discussion_r1252862820?
(You can also use the template tags in views by importing them `from django.templatetags.static import static`, see [SO](https://stackoverflow.com/questions/11721818/django-get-the-static-files-url-in-view))

integreat_compass/cms/templates/offers/image_picker.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n %}
-<div class="title-image-preview-container relative rounded rounded-3xl overflow-hidden {% if widget.value == default_title_image or not widget.value.url.strip %}hidden{% endif %}">
+<div class="title-image-preview-container relative rounded rounded-3xl overflow-hidden {% if not widget.value.url.strip %}hidden{% endif %}">
```
timobrembeck:
```suggestion
<div class="title-image-preview-container relative rounded rounded-3xl overflow-hidden {% if not widget.value %}hidden{% endif %}">
```
This is probably sufficient?

integreat_compass/cms/views/offers/offer_delete_view.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def form_valid(self, form, *args, **kwargs):
+        title = (
+            Offer.objects.filter(pk=self.request.POST.get("pk"))
+            .first()
+            .latest_version.title
+        )
+        messages.success(
+            self.request,
+            _('The offer "{}" has successfully been deleted.').format(title),
```
timobrembeck:
```suggestion
        messages.success(
            self.request,
            _('The offer "{}" has successfully been deleted.').format(
                self.object.latest_version.title
            ),
```
No need for an additional query, the base Django view already retrieves the offer from the database.

Apart from that: If you don't perform any additional checks, it's preferred to use `get()` in the first place instead of `filter().first()`, since if no such object exists, the error `Offer matching query does not exist` is more informative than `'NoneType' object has no attribute 'latest_version'`.

integreat_compass/cms/views/offers/offer_delete_view.py
```diff
@@ -0,0 +1,40 @@
...
+    View showing all of a user's offers
+    """
+
+    model = Offer
+    http_method_names = ["post"]
+    success_url = reverse_lazy("cms:protected:my_offers")
+
+    def get_permission_required(self):
+        offer_instance = Offer.objects.filter(pk=self.request.POST.get("pk")).first()
+        if offer_instance.creator != self.request.user:
```
timobrembeck:
```suggestion
        if self.get_object().creator != self.request.user:
```
Unfortunately, `self.object` is not yet populated at this point in time, but we could use the `self.get_object()` to retrieve it. Does not make a difference in sql queries though.

timobrembeck (changes requested):
Awesome, thanks! :+1:

I really like the increased usage of inbuilt Django views!

integreat_compass/cms/forms/offers/offer_version_form.py
charludo: `None` works (in the sense that no error is thrown), nut removing images stops working. I think `None` fields are just not passed to the model (?).

integreat_compass/cms/templates/offers/my_offers_list.html
charludo: Yep, good idea.

integreat_compass/cms/views/offers/offer_delete_view.py
charludo: Alright, thanks for letting me know!!

JoeyStk:
> @JoeyStk is there anything else you want to change?

Not really. Maybe just tests? Thank you in advance for your help on this PR :)

integreat_compass/cms/forms/offers/offer_version_form.py
timobrembeck: Ah ok I get it, the way how it is supposed to work is setting `checked = true` on the hidden `offer_version-title_image-clear_id` checkbox in JS when clearing the input. Then, we don't need any custom validation at all in the form class,

timobrembeck: Ah, I just noticed this is already the case? So then we can probably remove the custom clean method?

https://github.com/digitalfabrik/integreat-compass/blob/036de377e68ba5379375fb9ec9fe8563efabf0ce/integreat_compass/static/src/js/image_picker.ts#L26

charludo: Oh - indeed. Must have become superfluous at some point I guess :D

timobrembeck (approved):
Awesome, thanks! :rocket:

PR #73, opened 2023-06-28
Fix bug preventing submission of online-only offers

### LINKED ISSUE #72: Online-only offers can't be submitted
### Describe the Bug

When setting an offer to online-only, the Map and address field are hidden. However, the form will not send, since address and coordinates are required.
To make matter worse, this is not communicated to the user, since the fields are hidden.

### Steps to Reproduce

Create a new offer, set to "online" without filling in an address first. Try to submit.

### Expected Behavior

Form sends and saves

### Actual Behavior

Nothing happens

### Additional Information

Properly validating this might be tricky, since location and offer are handled in different forms/models

charludo:
### Short description

Fix bug preventing submission of online-only offers

### Proposed changes

- address, lat, long fields are no longer required
- in the LocationForm `clean` method, check if the offer mode is ONLINE; if not, and one of the mentioned fields is missing, show a message to the user

### Side effects

- none I am aware of

### Resolved issues

Fixes: #72

### DIFF
--- a/integreat_compass/cms/forms/offers/location_form.py
+++ b/integreat_compass/cms/forms/offers/location_form.py
@@ -1,7 +1,9 @@
 import logging
 
 from django import forms
+from django.utils.translation import gettext_lazy as _
 
+from ...constants import offer_mode_types
 from ...models import Location
 from ..custom_model_form import CustomModelForm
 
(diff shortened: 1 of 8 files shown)


JoeyStk (commented):
Thank you for catching this bug. ~~However for me this somehow still didn't work. I logged in as root and tried to submit a offer only for online, but the form wouldn't get send. Also I didn't receive a warning message. For me it still looks the same as on develop branch. I have emptied the cache, but still it wouldn't work. Do you have a different idea what I could try?~~ (this was an issue on my side, I was able to fix it now). However I still do get an error when submitting the form, but it is a different one than before. Now I get an IntegrityError (I can send you a screenshot if needed).

Additionally I also found something that isn't really part of this PR, but since it's also related to the application form, I thought it might by okay to report it here. There is a typo in the [help text for phone number](https://github.com/digitalfabrik/integreat-compass/blob/develop/integreat_compass/cms/models/organizations/contact.py#L27). Could you please fix that too? :)

charludo:
> I can send you a screenshot if needed

Please do!

JoeyStk (approved):
I was finally fix the issue on my side and test the PR properly. Thanks for fixing the typo. The other things work as expected

PR #75, opened 2023-07-02
Add view of all offers

### LINKED ISSUE #39: Search & Filter available offers
### Motivation
Core functionality of the tool.

### User Story
As a job center counselor I want filter/search all available offers so that I can find applicable and high quality offers for my clients.

### Additional Context
- This should be the default view after logging in an for unauthenticated users.
- Unauthenticated users should also allowed to see and search offers.

### Screenshots
image

### LINKED ISSUE #40: View offer details

JoeyStk:
### Short description
This PR adds the view to see all offers at once. This view also provides the option to search and filter the provided offers.

### Proposed changes

- Remove the current index view, since it's no longer wanted
- Instead add the all offers view as a replacement
- Show all offers
- Add functionality to filter and search the offers
- On click on one of the offers open a detail view for the offer

### Side effects
This PR is still a work in progress. Please find the questions, that made me get stuck in the comments below

### Resolved issues

Fixes: #39
Fixes: #40


integreat_compass/cms/views/index.py
```diff
@@ -1,9 +1,124 @@
...
+        }
+        
+        filtered_offers = []
+        selected_filters["tags"] = [eval(i) for i in selected_filters["tags"]]
+
+
+        for offer in offers:
+            tag_ids = [tag.id for tag in offer.tags.all()]
+
+            if any(item in selected_filters["tags"] for item in tag_ids):
```
JoeyStk: This part is supposed to do the logic for filtering. However it's not fully functional yet - and also doesn't seem to be the best solution at all. All parts work individually, but they don't work together. Do you have an idea, how to create the filter logic if two or more filter criteria have been selected (for example online offer and has "Beruf" in the title)

integreat_compass/cms/views/index.py
```diff
@@ -1,9 +1,124 @@
...
+        """
+        offer_dict = {}
+        offer_dict["image"] = offer.versions.first().title_image.url
+        offer_dict["title"] = offer.versions.first().title
+        offer_dict["organization"] = offer.organization.name
+        offer_dict["description"] = offer.versions.first().description
+        offer_dict["creation_date"] = offer.versions.first().offer_version_date
+        offer_dict["tags"] = offer.tags.all()
+        return offer_dict
+
```
JoeyStk: With the current implementation we would need also a "reset filter" button. I don't mind that, but if there is a better solution it would be nice if we could circumvent it

tests/cms/test_index.py
```diff
@@ -0,0 +1,10 @@
+import pytest
```
JoeyStk, line 1: Please ignore the test / this file. I'm still trying to wrap my head around the required tests

integreat_compass/cms/views/index.py
timobrembeck: When filtering database objects in Django, it's usually best to do the filtering via the [queryset filters](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#filter).
This has the advantage that Django can [defer the execution](https://docs.djangoproject.com/en/4.2/topics/db/queries/#querysets-are-lazy) of the sql statements and only execute one query in the end which contains the final objects, so we don't retrieve objects from the database which we don't need anyway.

The only exception to this rule is when you e.g. need the intermediary results or perform other operations which intercept the lazy loading and cause multiple queries.

In this case, it depends on how the filtering should work - if multiple tags are selected, do we want all offers which contain at least one of these tags, or all of them?

The first one is a bit easier:
```
offers = offers.filter(tags__in=selected_filters["tags"]).distinct()
```

While the latter requires a loop (but due to the lazy execution of querysets, this only generates one query in the end):
```
for tag in selected_filters["tags"]:
    offers = offers.filter(tags=tag)
```

And this loop can also be written as fancy python syntax (but this only changes way the query is generated, it does not change the fact that the filtering happens on the sql layer):
```
from functools import reduce

offers = reduce(lambda offers, tag: offers.filter(tags=tag), selected_filters["tags"], offers)
```

In any case, you don't need to know the whole queryset documentation by heart, but you should know where to look whenever you face similar tasks in the future.

integreat_compass/cms/views/index.py
```diff
@@ -1,9 +1,125 @@
...
+        selected_filters = {
+            "tags": request.POST.getlist("offer-tags"),
+            "mode_type": request.POST.get("mode-type"),
+            "group_type": request.POST.get("group-type"),
+            "search": request.POST.get("offer-search"),
+        }
+
+        filtered_offers = []
+        selected_filters["tags"] = [literal_eval(i) for i in selected_filters["tags"]]
```
timobrembeck: Even if `literal_eval` is less scary than the evil `eval()`, it's not a function you should use often, since the input is completely user controlled and if you expect the input to have a certain type (in this case an `int`), you might encounter strange edge cases in your code, when the input suddenly is a completely different Python data structure.

Anyways, in this case the problem you are trying to solve is "I get string as input and want integers as output", so I don't really understand how you're ending up with `literal_eval()` as solution instead of just `int()`. Also, Django already provides a mechanism for this task: forms. If you want to have input forms and retrieve the input data as Python types other than strings, please create a form class, define the field as `IntegerField`, `ForeignKey`, etc. and access `form.cleaned_data` to get the user input as python data structure.

Apart from that, you don't even need to do this, since Django's queryset filters also work with string input, e.g. `Offer.objects.filter(tags="2")` is totally fine.

integreat_compass/cms/views/index.py
```diff
@@ -1,9 +1,125 @@
...
 
 class IndexView(TemplateView):
     """
-    Start page of the project
+    View to see all offers
     """
 
     template_name = "index.html"
+
+    def get(self, request, *args, **kwargs):
```
timobrembeck: If you don't intercept the rendering process itself, but just add additional context data, please just override `get_context_data()` instead of `get()`.

integreat_compass/cms/views/index.py
```diff
@@ -1,9 +1,125 @@
...
+            {
+                "offers": filtered_offers,
+                "tags": tags,
+                "amount_of_offers": amount_of_offers,
+                "offer_group_types": offer_group_types.CHOICES,
+                "offer_mode_types": offer_mode_types.CHOICES,
+            },
+        )
+
+    def store_offer_data_in_dict(self, offer):
```
timobrembeck: I don't really understand the purpose of this function. You just can access all of these fields directly in the template, I think if you're renaming the variables you just make it harder to get where a variable is coming from.
Also, accessing `offer.versions.first()` multiple times generates a lot of unnecessary queries, please add a new cached property to the offer model to remember this field the first time you're accessing it.

integreat_compass/static/src/index.ts
```diff
@@ -4,9 +4,12 @@ import "./js/display_filenames";
 import "./js/image_picker";
 import "./js/map";
 import { createIconsAt } from "./js/utils/create-icons";
+import { handleOfferDetails, countLengthOfReview, setLengthOfReview } from "./js/offer-details";
```
timobrembeck: Did you forget to push `./js/offer-details.js`?

integreat_compass/locale/de/LC_MESSAGES/django.po
```diff
@@ -449,12 +449,39 @@ msgid "Home"
...
+#: cms/templates/index.html
+#, python-format
+msgid ""
+"\n"
+"                        %(amount_of_offers)s offers\n"
+"                        "
+msgstr ""
+"\n"
+"                        %(amount_of_offers)s Maßnahmen\n"
+"                        "
```
timobrembeck: In order to avoid these weird spaces, please use `trimmed` on the translation tag.

integreat_compass/cms/views/index.py
```diff
@@ -1,9 +1,125 @@
...
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        offers = Offer.objects.all()
+        tags = Tag.objects.all()
+        list_of_offers = []
+        for offer in offers:
+            list_of_offers.append(self.store_offer_data_in_dict(offer))
+        amount_of_offers = len(list_of_offers)
```
timobrembeck: You don't need to calculate this in the view, you can just call the `len` template filter in the template.

integreat_compass/cms/views/index.py
```diff
@@ -1,9 +1,125 @@
+from ast import literal_eval
+
+from django.shortcuts import render
 from django.views.generic import TemplateView
 
+from ..constants import offer_group_types, offer_mode_types
+from ..models import Offer, Tag
+
 
 class IndexView(TemplateView):
```
timobrembeck: Also, while you're at it, in my opinion it would make sense to convert this to a Django [`ListView`](https://docs.djangoproject.com/en/4.2/ref/class-based-views/generic-display/#listview) to make use of a bit more inbuilt functionality...

integreat_compass/static/src/index.ts
JoeyStk: I wanted to do it in two PRs, but I have changed my mind now and will do it in one PR. The according files are uploaded now as well. Sorry for the TS error

integreat_compass/cms/views/index.py
JoeyStk: Thank you very much for this detailed description. Turns out I confused myself just before asking the question. The bigger issue for me was how to stack filters on top of each other, but turns out we did it for Integreat too and it's quite simple. Sorry!

integreat_compass/cms/views/index.py
JoeyStk: Thank you for explaining, but this resolved itself since I switched to the Django filters :) I will keep it in mind for the future!

integreat_compass/cms/views/index.py
JoeyStk: This was a mindblow! Thanks ;) Is applied now

integreat_compass/locale/de/LC_MESSAGES/django.po
JoeyStk: Thanks! This also resolved itself, but that was very helpful for other PRs!

integreat_compass/cms/views/index.py
JoeyStk: This should be applied now :)

integreat_compass/cms/views/index.py
JoeyStk: This is not resolved yet, I still need to some research on it

integreat_compass/cms/templates/_base.html
```diff
@@ -11,26 +11,27 @@
                      alt="{% translate "Logo" %}"
                      class="h-24 w-48">
             </a>
+            <a href="{% url 'cms:public:index' %}"
+               class="ml-6 mr-4 hover:text-gray-900">{% translate "All offers" %}</a>
             <a href="#" class="ml-6 mr-4 hover:text-gray-900">{% translate "Dashboard" %}</a>
```
charludo: We don't need the dashboard link anymore, right? Since it is functionally identical to "All Offers" now. (Which in turn does the same as clocking the logo 😅 😄 )

integreat_compass/cms/templates/index.html
```diff
@@ -4,15 +4,114 @@
...
-    <div class="absolute top-0 right-0 bottom-0 left-0 flex justify-center items-center bg-background-900 z-1">
-        <div class="flex flex-col justify-center gap-4 text-center">
-            <h1 class="font-bold text-5xl drop-shadow-md mb-12 text-white">{% translate "Arriving, reimagined." %}</h1>
-            <a href="#"
-               class="bg-primary hover:bg-red-500 text-white font-bold text-3xl py-4 px-4 rounded-2xl">{% translate "Discover Integreat Compass" %}</a>
-            <a href="{% url 'cms:protected:new_offer' %}"
-               class="bg-white hover:bg-gray-200 font-bold text-3xl py-4 px-4 rounded-2xl">{% translate "Add new offer" %}</a>
+    <div class="flex container mx-auto h-[calc(100vh-6rem)]">
+        <div class="hidden lg:block fixed top-24 bottom-0 bg-white lg:w-3/12 py-8 pr-8 shadow-[10px_0_5px_-2px_rgba(0,0,0,0.1)]">
+            <form action="{% url 'cms:public:index' %}" method="post">
```
charludo: On wider screen I have a large empty space to the left of the filter bar (due to the `max-width` of the container). That is intentional for most pages, however for this specific view I think it does not look good, since we already have a two column layout. 

We could just remove the left margin for this view, or maybe Iabella has a good idea?

image

integreat_compass/cms/templates/index.html
```diff
@@ -4,15 +4,114 @@
...
 {% endblock title %}
 {% block content %}
-    <div class="absolute top-0 right-0 bottom-0 left-0 flex justify-center items-center bg-background-900 z-1">
-        <div class="flex flex-col justify-center gap-4 text-center">
-            <h1 class="font-bold text-5xl drop-shadow-md mb-12 text-white">{% translate "Arriving, reimagined." %}</h1>
-            <a href="#"
-               class="bg-primary hover:bg-red-500 text-white font-bold text-3xl py-4 px-4 rounded-2xl">{% translate "Discover Integreat Compass" %}</a>
-            <a href="{% url 'cms:protected:new_offer' %}"
-               class="bg-white hover:bg-gray-200 font-bold text-3xl py-4 px-4 rounded-2xl">{% translate "Add new offer" %}</a>
+    <div class="flex container mx-auto h-[calc(100vh-6rem)]">
```
charludo: Neither `mx-auto` nor the height calculation seem to have any effect

integreat_compass/cms/templates/index.html
```diff
@@ -4,15 +4,114 @@
...
+                                               name="mode-type"
+                                               value="{{ mode_type.0 }}">
+                                        <label for="mode-type-{{ mode_type.0 }}"
+                                               class="w-full duration-100 m-0 p-2 text-lg rounded-full peer-checked:bg-primary peer-checked:text-white">
+                                            {{ mode_type.1 }}
+                                        </label>
+                                    </div>
+                                </li>
+                            {% endfor %}
+                        </ul>
```
charludo: I know it's in the designs like this, but IMO these should definitely exhibit checkbox behavior, not radio button behavior (e.g. I want to filter for all offers that are either online or hybrid).

integreat_compass/cms/templates/index.html
```diff
@@ -4,15 +4,114 @@
...
+                    </div>
+                    <div class="mt-4">
+                        <button class="btn" type="submit">{% translate "Submit" %}</button>
```
charludo: If we decided to do it this way and I just forgot, ignore the following - 
But IMO filters should never have to be submitted. These should either filter in JS, or request the filtered offers via ajax calls.

Also I think from an Integration Specialist perspective it is imperative that filters are reflected in the URL (as GET parameteres, rahtern than POST), so filtered views can be bookmarked, shared,...

integreat_compass/cms/templates/index.html
```diff
@@ -4,15 +4,114 @@
...
+                                       name="offer-tags"
+                                       class="hidden peer"
+                                       type="checkbox"
+                                       value="{{ tag.id }}">
+                                <label for="{{ tag.id }}"
+                                       class="px-4 py-2 text-lg bg-white rounded-full text-center shadow-lg peer-checked:bg-primary peer-checked:text-white peer-checked:border-none">
+                                    {{ tag }}
+                                </label>
+                            </div>
+                        {% endfor %}
```
charludo: (This isn't really a comment on the PR, but I think it's as good a time to ask as any - ) What is the intended behavior for these? All tags unclicked -> Show all, at least one clicked -> show all matching at least one (or all?)

integreat_compass/static/src/js/offer-details.ts
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+};
+
+export const setLengthOfReview = (length: unknown) => {
+    const offerRatingUsedWords = document.getElementById("offer-rating-used-words");
+    if (offerRatingUsedWords) {
+        offerRatingUsedWords.textContent = length as string;
+    }
+};
```
charludo: Why use `unknown` instead of `number` here?

charludo (changes requested):
Thank you for this (draft) PR! This is quite a big step towards the MVP! 🎉 

I have some nitpicks + some general usability concerns:

integreat_compass/cms/templates/_base.html
JoeyStk: Good question. I think you're more up-to-date about this than I am, but I thought the dashboard is a customized view, that depending on your role shows different content. But yes, this means, that at least for not logged in user roles this view is obsolete now. My question is: Should I remove it altogether or rather just make it visible for logged in users?

integreat_compass/cms/templates/index.html
JoeyStk: Good question. I will raise this question on Wednesday

integreat_compass/cms/templates/_base.html
charludo: Oh, interesting, so basically the dashboard would be:
- not logged in: All Offers
- board member: Applications to vote on
- integration specialist: All offers
- offer provider: My offers (?)
I guess that makes *some* sense. There is a dedicated Dashboard mockup though:
image
but it's not part of the MVP. So I was under the impression that this would, eventually, become the dashboard for all logged in users, and that until then Dashboard = All Offers.

JoeyStk: Yes, exactly that are the versions of the dashboard as I understood them too. I didn't realize however that this is not part of the MVP :thinking: Do you agree that for now it would be best to remove this point from the navigation or would you just redirect from Dashboard to All Offers?

integreat_compass/cms/templates/index.html
JoeyStk: I will also ask this on Wednesday  :)

integreat_compass/cms/templates/index.html
JoeyStk: Oh, you're right. I remember that at one point of the process I added it out of necessity, but since then I made some change that has made this piece of code obsolete

integreat_compass/cms/templates/index.html
JoeyStk: One way I was able to make it work was by playing with fire and adding negative margin to the sidebar on the left and adding the same spacing as (positive) padding to the form. With this the spacing evens out and we have the expected result 
grafik
However this is not a great solution since the code is really messy (3xl:ml-[calc(-1*(calc(calc(100vw-1200px)/2)))] lg:ml-[calc(-1*(calc(calc(100vw-1024px)/2)))]) and for the form element (3xl:ml-[calc(calc(100vw-1200px)/2)] lg:ml-[calc(calc(100vw-1024px)/2)]) and still there is no guarantee that it will work for every width (especially ultra large screens will look awful). 

> We could just remove the left margin for this view

I think we would need to remove the mx-auto from the base view and then instead add it to every template that needs it. To me that is also not a great solution. Does someone maybe have another idea?

timobrembeck: > Does someone maybe have another idea?

Instead of `<div class="container">`, you could write `<div class="{% block container_width %}max-w-... md:max-w-... etc.{% endblock container_width %}">` in the base template, and then in the `offer-detail-layover.html` template, you could overwrite this by adding `{% block container_width %}max-w-... md:max-w-... etc.{% endblock container_width %}` with the different width styles.

integreat_compass/cms/templates/index.html
JoeyStk: Yes, it was agreed on today, that we should build this as checkboxes and remove the option "both" for group or alone

integreat_compass/cms/templates/index.html
JoeyStk: It was agree to leave it as it is. All tags unclicked -> Show all, at least one clicked -> show all matching all (decided against at least one)

integreat_compass/cms/templates/_base.html
charludo: I think I'd rather remove it for now, then later we can add it as a new "feature" instead of changing what the link does for the users

JoeyStk: Is the dashboard not part of the MVP?

integreat_compass/cms/templates/index.html
JoeyStk: I think we need the height calculation to compensate the height of the header and not have the scrollbar on the right side of the screen. I moved the height calculation to the body now, so it's working again
Before:
Screenshot from 2023-07-26 11-15-12
After: 
Screenshot from 2023-07-26 11-15-39

integreat_compass/cms/templates/_base.html
svenseeberg: We discussed this, I forgot the result. But I think we said that we do not want to implement the Dashboard in the MVP and defined the list + filters as the main page?

JoeyStk:
@charludo and @svenseeberg please feel free to check out the failing tests, especially the last test in pytest

integreat_compass/cms/templates/index.html
JoeyStk: Should be implemented now :)

integreat_compass/cms/templates/index.html
JoeyStk: Thank you for your suggestion. I took the main part, but changed the details a bit, I hope the container works as espected

integreat_compass/cms/templates/_base.html
JoeyStk: The dashboard was removed as discussed

integreat_compass/cms/views/index.py
JoeyStk: I changed this function to the get_context_data function, but later added the get function for the filter form

integreat_compass/cms/views/index.py
```diff
@@ -1,9 +1,93 @@
...
+            {
+                "tags": tags,
+                "offers": filtered_offers,
+                "filtered_tags": request.GET.getlist("offer-tags"),
+                "offer_group_types": offer_group_types.CHOICES,
+                "filtered_group_types": request.GET.getlist("group-type"),
+                "offer_mode_types": offer_mode_types.CHOICES,
+                "filtered_mode_types": request.GET.getlist("mode-type"),
+            },
+        )
```
charludo:
```suggestion
        filtered_offers = Offer.objects.all()
        if offer_tags := request.GET.getlist("offer-tags"):
            filtered_offers = filtered_offers.filter(tags__in=offer_tags)

        if mode_types := request.GET.getlist("mode-type"):
            filtered_offers = filtered_offers.filter(mode_type__in=mode_types)

        if group_types := request.GET.getlist("group-type"):
            filtered_offers = filtered_offers.filter(group_type__in=group_types)

        if offer_search := request.GET.get("offer-search"):
            filtered_offers = filtered_offers.filter(
                versions__title__icontains=offer_search
            )

        return render(
            request,
            self.template_name,
            {
                "tags": Tag.objects.all(),
                "offers": filtered_offers.distinct(),
                "filtered_tags": offer_tags,
                "offer_group_types": offer_group_types.CHOICES,
                "filtered_group_types": group_types,
                "offer_mode_types": offer_mode_types.CHOICES,
                "filtered_mode_types": mode_types,
            },
        )
```

This is a suggestion for making the code a little more readable. At the same time, I am having an issue with the (current!) implementation.

For example, click "Berufssprache" -> filter works correctly.
Then click "Gesellschaft" -> only "Deutsch für Anfänger" is shown.

This makes no sense, since either both should be shown (offers matching at least one tag), or none (offers matching all tags).

I do not really know what the reason is.

The suggested code implements the "matching at least one tag" tag. I could not get the "matching all tags" to work, unsure what the reason is.

integreat_compass/cms/templates/index.html
```diff
@@ -3,16 +3,128 @@
...
+                                    {% if offer.versions.first.title_image %}
+                                        <img class="rounded-2xl"
+                                             src="{{ offer.versions.first.title_image.url }}"
+                                             width="100%"
+                                             height="100%"
+                                             alt="">
+                                    {% endif %}
+                                </div>
+                                <div class="w-8/12 py-2 px-4">
+                                    <h3 class="font-bold">{{ offer.versions.first.title }}</h3>
```
charludo:
```suggestion
                                    <h3 class="font-bold">{{ offer.public_version.title }}</h3>
```
Sorry I did not notice this before. This has to be done everywhere you use any property of the offer version, since users should never see non-public version content.

integreat_compass/cms/views/index.py
```diff
@@ -1,9 +1,93 @@
...
+        :type args: list
+
+        :param kwargs: The supplied keyword arguments
+        :type kwargs: dict
+
+        :return: Response for filtered offers
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        tags = Tag.objects.all()
+        filtered_offers = Offer.objects.all()
```
charludo, line 59:
```suggestion
        filtered_offers = Offer.objects.exclude(public_version__isnull=True)
```
Something along these lines is necessary so we do not attempt to show new/non-public offers

charludo (changes requested):
Thanks! This is looking pretty good now.

Unfortunately, I have some additional thoughts now 😅 
- We should probably not pass the CSRF token as a GET parameter, if only for the reason that it makes the link a bit ugly, and also probably not shareable (!)
- we could shorten the filter names for the sake of a more readable URL: `offer-tags` -> `tags`, `offer-search` -> `search` or `query` and maybe also `mode-type` -> `mode` and `group-type` -> `group`  (this is fully optional though, feel free to postpone to a later PR/issue)
- please rename `offer-detail-layover.html` to `offer-detail-overlay.html`

Super sorry for these new change requests today, feel free to tell me if you don't have time or want help!

charludo:
Just noticed: It would probably be a good idea to create two additional example offers, approve them, and add them to the test data fixture (since otherwise, the list of public offers is empty)

JoeyStk:
Thank you for your feedback. All raised issues should be resolved, if I didn't mess up the rebase again :see_no_evil:. The only issue that's left is [this line](https://github.com/digitalfabrik/integreat-compass/pull/75/files#diff-71953a6d8fb8de162bf0c99091c3142dbd8256d8003ef2fc1fd23b4db14aacf1R59) that raises a Django error "Cannot resolve keyword 'public_version' into field". @charludo could you please have another look at it? :)

charludo:
> Thank you for your feedback. All raised issues should be resolved, if I didn't mess up the rebase again see_no_evil. The only issue that's left is [this line](https://github.com/digitalfabrik/integreat-compass/pull/75/files#diff-71953a6d8fb8de162bf0c99091c3142dbd8256d8003ef2fc1fd23b4db14aacf1R59) that raises a Django error "Cannot resolve keyword 'public_version' into field". @charludo could you please have another look at it? :)

You're right, it appears filtering by model property is not possible, sadly. I opted for a list comprehension after we are done working on the queryset - @timoludwig, do you have a smart idea how to do this more nicely? (In the mean time though, I guess this works...)

Also added some new test data and modified the tests accordingly.

From my side, this is ready to merge, even if we come back to some details later.

charludo (approved):

svenseeberg (approved):

timobrembeck:
> You're right, it appears filtering by model property is not possible, sadly. I opted for a list comprehension after we are done working on the queryset - @timoludwig, do you have a smart idea how to do this more nicely?

The only way to do this is to "re-implement" the property via queryset methods. Since the possibilities of Django's `Q()` & `F()` expressions, aggregations and annotations are nearly endless, I'm pretty confident that for almost every python property, there is a way to calculate it on the sql layer. It's just a lot more complicated to write these complex queries than to express the same thing in simple python code.

In this particular case however, I don't think it's too complicated - since PR #70 added the change that the `state` field of offer versions is stored in the database instead of calculated dynamically. If I understood correctly, we want to filter for `Offer` objects that have at least one approved version? This could easily be expressed as:
```
filtered_offers = filtered_offers.filter(versions__state=offer_version_states.APPROVED).distinct()
```
(hopefully - I didn't test this, but to me it looks like it should work :innocent:)

charludo:
> ```
> filtered_offers = filtered_offers.filter(versions__state=offer_version_states.APPROVED).distinct()
> ```
> 
> (hopefully - I didn't test this, but to me it looks like it should work innocent)

Ahh, good idea, I didn't even think about using the state, but you are right of course, this implies the existence of a public version. Oh well, now it's merged, so let's put this on the "later" pile 😅

PR #76, opened 2023-07-05
Add offer provider registration form

### LINKED ISSUE #52: Register "Offer Provider"
### Motivation
NGOs should be able to freely create applications for offers.

### User Story
As a "Bildungsanbieter" I want create an account so that I can create offers. 

### Proposed Solution
- User registration form with e-mail validation
- Default role "Bildungsanbieter" (allowed to create offers). We can also define as "no role" to be the lowest permission level and that these users are allowed to create offer applications.
- First name and last name are required.

### Alternatives
- Manually create accounts

### Additional Context
- A user is not related to any organization (can freely enter offer applications for any organization).

charludo:
### Short description

Add a form (and needed logic) for users to register themselves as an offer provider

### Proposed changes

- add the form
- send an activation link
- activate account on clicking that link

### Side effects

- none I am aware of

### Resolved issues

Fixes: #52

### DIFF
--- a/integreat_compass/cms/forms/registration_form.py
+++ b/integreat_compass/cms/forms/registration_form.py
@@ -1,0 +1,111 @@
+import logging
+
+from django import forms
+from django.contrib.auth.password_validation import validate_password
+from django.utils.translation import gettext as _
+
+from ..constants import group_names
+from ..models import User
+from .custom_model_form import CustomModelForm
+
+logger = logging.getLogger(__name__)
+
+
+class RegistrationForm(CustomModelForm):
+    """
+    Form for user self-registration as offer providers
+    """
+
+    password = forms.CharField(
+        widget=forms.PasswordInput(), validators=[validate_password]
+    )
+    password_confirm = forms.CharField(
+        label=_("Confirm password"), widget=forms.PasswordInput()
+    )
+
+    class Meta:
+        """
+        This class contains additional meta configuration of the form class, see the :class:`django.forms.ModelForm`
+        for more information.
+        """
+
+        model = User
+        fields = ["email", "display_name"]
+
+    def clean(self):
+        """
+        Validate form fields which depend on each other, see :meth:`django.forms.Form.clean`:
+        If passwords do not match, add a :class:`~django.core.exceptions.ValidationError`.
+
+        :return: The cleaned form data
+        :rtype: dict
+        """
+        cleaned_data = super().clean()
+        if cleaned_data.get("password") != cleaned_data.get("password_confirm"):
+            self.add_error(
+                "password_confirm", forms.ValidationError(_("Passwords do not match."))
+            )
+        return cleaned_data
+
+    def clean_email(self):
+        """
+        Ensure the email does not exist yet, see :ref:`overriding-modelform-clean-method`:
+        If the email is already registered, add a :class:`~django.core.exceptions.ValidationError`.
+
+        :return: The password
+        :rtype: str
+        """
+
+        email = self.cleaned_data["email"]
+        if User.objects.filter(email=email).first():
+            self.add_error(
+                "email",
+                forms.ValidationError(
+                    _(
+                        'An account for the email "{}" does already exist. Did you mean to log in?'
+                    ).format(email)
+                ),
+            )
+        return email
+
+    def clean_display_name(self):
+        """
+        Ensure the display_name does not exist yet, see :ref:`overriding-modelform-clean-method`:
+        If the display_name is already taken, add a :class:`~django.core.exceptions.ValidationError`.
+
+        :return: The display_name
+        :rtype: str
+        """
+
+        display_name = self.cleaned_data["display_name"]
+        if User.objects.filter(display_name=display_name).first():
+            self.add_error(
+                "display_name",
+                forms.ValidationError(
+                    _('The display name "{}" is already taken.').format(display_name)
+                ),
+            )
+        return display_name
+
+    def save(self, commit=True):
+        """
+        This method extends the default ``save()``-method of the base :class:`~django.forms.ModelForm`
+        to create a new non-active user with role Offer Provider.
+
+        :param commit: Whether or not the changes should be written to the database
+        :type commit: bool
+
+        :return: The saved region object
+        :rtype: ~integreat_cms.cms.models.offers.offer_version.OfferVersion
+        """
+        cleaned_data = self.cleaned_data
+        new_user = User.objects.create_user(
+            email=cleaned_data["email"],
+            display_name=cleaned_data["display_name"],
+            password=cleaned_data["password"],
+            group=group_names.OFFER_PROVIDER,
+            is_active=False,
+        )
+
+        logger.debug("Created new offer provider %s", self.cleaned_data["email"])
+        return new_user
--- a/integreat_compass/cms/views/authentication/registration_view.py
+++ b/integreat_compass/cms/views/authentication/registration_view.py
@@ -1,0 +1,39 @@
+import logging
+
+from django.contrib import messages
+from django.urls import reverse_lazy
+from django.utils.translation import gettext as _
+from django.views.generic.edit import CreateView
+
+from ...forms import RegistrationForm
+from ...utils.email_utils import send_activation_mail
+
+logger = logging.getLogger(__name__)
+
+
+class RegistrationView(CreateView):
+    """
+    View allowing new users to sign up as offer providers
+    """
+
+    template_name = "authentication/registration.html"
+    form_class = RegistrationForm
+    success_url = reverse_lazy("cms:public:login")
+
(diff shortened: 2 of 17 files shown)


integreat_compass/locale/de/LC_MESSAGES/django.po
```diff
@@ -448,6 +471,45 @@ msgstr "Passwort eingeben"
...
+msgid ""
+"You are receiving this e-mail because you created an account with Integreat "
+"Compass."
+msgstr ""
+"Sie erhalten diese Email da Sie einen Account im Integreat Compass angelegt "
+"haben."
+
+#: cms/templates/emails/activation_email.html
+#: cms/templates/emails/activation_email.txt
+msgid "Please click the link below to activate your account."
```
svenseeberg: The tech writer in my would write `Please use the link below` to avoid being to specific about the input device. It would also work with a keyboard, etc.

integreat_compass/locale/de/LC_MESSAGES/django.po
```diff
@@ -448,6 +471,45 @@ msgstr "Passwort eingeben"
...
+msgstr "Klicken Sie den folgenden Link um Ihren Account zu aktivieren."
+
+#: cms/templates/emails/activation_email.txt
+msgid ""
+"If clicking on the link above does not work, please copy the URL and paste "
```
svenseeberg: Same here: `using` instead of `clicking`.

integreat_compass/locale/de/LC_MESSAGES/django.po
```diff
@@ -505,6 +567,49 @@ msgstr "Angaben der Organisation"
...
+#: cms/views/authentication/account_activation_view.py
+msgid ""
+"Please contact an administrator to request a new link to activate your "
+"account."
+msgstr ""
+"Bitte kontaktieren Sie einen Administrator um einen neuen Link anzufordern."
+
+#: cms/views/authentication/registration_view.py
+msgid ""
+"A confirmation email has been sent to \"{}\". Please click the link in the "
```
svenseeberg: Also replace `click` with `use`.

integreat_compass/core/settings.py
```diff
@@ -309,3 +309,59 @@
...
+#: Username to use for the SMTP server defined in :attr:`~integreat_compass.core.settings.EMAIL_HOST`
+#: (see :setting:`django:EMAIL_HOST_USER`). If empty, Django won’t attempt authentication.
+EMAIL_HOST_USER = os.environ.get("INTEGREAT_COMPASS_EMAIL_HOST_USER", SERVER_EMAIL)
+
+#: Port to use for the SMTP server defined in :attr:`~integreat_compass.core.settings.EMAIL_HOST`
+#: (see :setting:`django:EMAIL_PORT`)
+EMAIL_PORT = int(os.environ.get("INTEGREAT_COMPASS_EMAIL_PORT", 587))
+
+#: Whether to use a TLS (secure) connection when talking to the SMTP server.
+#: This is used for explicit TLS connections, generally on port 587.
```
svenseeberg: This is copied from the Integreat settings.py, right? IMHO a better description would be `This will use Opportunistic TLS (STARTTLS command after starting a plain text connection).`

integreat_compass/core/settings.py
```diff
@@ -309,3 +309,59 @@
...
+EMAIL_BACKEND = (
+    "django.core.mail.backends.console.EmailBackend"
+    if DEBUG
+    else "django.core.mail.backends.smtp.EmailBackend"
+)
+
+#: Default email address to use for various automated correspondence from the site manager(s)
+#: (see :setting:`django:DEFAULT_FROM_EMAIL`)
+DEFAULT_FROM_EMAIL = os.environ.get(
+    "INTEGREAT_COMPASS_SERVER_EMAIL", "keineantwort@integreat-compass.de"
```
svenseeberg: I personally would only use @example.com mail addresses in public code. But that is just a personal preference ^^

svenseeberg (commented):
A non-pythonic review ;-)

integreat_compass/cms/views/authentication/registration_view.py
```diff
@@ -0,0 +1,87 @@
...
+from django.utils.translation import gettext as _
+from django.views.generic import TemplateView
+
+from ...forms import RegistrationForm
+from ...utils.email_utils import send_activation_mail
+
+logger = logging.getLogger(__name__)
+
+
+class RegistrationView(TemplateView):
```
timobrembeck: In my opinion, it would make sense to inherit from [`CreateView`](https://docs.djangoproject.com/en/4.2/ref/class-based-views/generic-editing/#django.views.generic.edit.CreateView) here.

integreat_compass/cms/forms/registration_form.py
```diff
@@ -0,0 +1,111 @@
...
+from django.utils.translation import gettext as _
+
+from ..constants import group_names
+from ..models import User
+from .custom_model_form import CustomModelForm
+
+logger = logging.getLogger(__name__)
+
+
+class RegistrationForm(CustomModelForm):
```
timobrembeck, line 14: I think it makes sense to inherit from [`UserCreationForm`](https://docs.djangoproject.com/en/4.2/topics/auth/default/#django.contrib.auth.forms.UserCreationForm) here.

integreat_compass/core/settings.py
```diff
@@ -309,3 +309,59 @@
...
+
+#: Whether to use a TLS (secure) connection when talking to the SMTP server.
+#: This is used for explicit TLS connections, generally on port 587.
+#: (see :setting:`django:EMAIL_USE_TLS`)
+EMAIL_USE_TLS = bool(
+    strtobool(os.environ.get("INTEGREAT_COMPASS_EMAIL_USE_TLS", "True"))
+)
+
+#: Whether to use an implicit TLS (secure) connection when talking to the SMTP server.
+#: In most email documentation this type of TLS connection is referred to as SSL. It is generally used on port 465.
```
timobrembeck, line 361: @svenseeberg Apparently, port 465 seems to be deprecated, but does this mean SSL/TLS should not be used as well, or can it also be used via port 587? Or is STARTTLS via port 587 the only sensible option nowadays and SSL/TLS should be avoided altogether? In my mind, STARTTLS seems to be a bit inferior, since it sends the mail unencrypted if encryption fails?

integreat_compass/core/settings.py
timobrembeck: The `keineantwort@` is probably irrelevant if it does not have an inbox, but `tech@` could become subject to spam? However, for integreat-cms, this does not seem to be the case...

integreat_compass/cms/forms/registration_form.py
charludo: That was the plan initially, however this did not play nice with our user model (no usernames, mandatory email)

integreat_compass/core/settings.py
svenseeberg: I think (most) servers abort the connection if STARTTTLS fails. But I agree that it seems better to encrypt the traffic to begin with. Standards for some reason seem to disagree. But 465 will probably be supported for a long time to come.

svenseeberg: Oh interesting. https://datatracker.ietf.org/doc/html/rfc8314 and more interesting https://datatracker.ietf.org/doc/html/rfc8314#appendix-A. To sum things up: 465 with implicit TLS seems to be the future. For obvious reasons.

timobrembeck: Interesting, thanks for the insight! Especially weird that a lot of these journalistic/blog-type explanation websites state that 465 is deprecated and 587 should be preferred, so probably they didn't do their research properly or just used outdated sources which predate the RFC you linked.

Back to the PR: Should we derive the default email port from the used method? Something along the lines of
```
if EMAIL_USE_TLS:
    EMAIL_DEFAULT_PORT = 587
elif EMAIL_USE_SSL:
    EMAIL_DEFAULT_PORT = 465
else:
    EMAIL_DEFAULT_PORT = 25

EMAIL_PORT = int(os.environ.get("INTEGREAT_COMPASS_EMAIL_PORT", EMAIL_DEFAULT_PORT))
```
?

integreat_compass/cms/forms/registration_form.py
timobrembeck: Ah ok, I see. I guess the `BaseUserCreationForm` could still be used since it does not handle the username (and we can still extend the email and display name validation), but ok we don't save that much code, just a few lines of password handling...

timobrembeck (approved):
:tada:

integreat_compass/core/settings.py
svenseeberg: I think this looks reasonable, yes.

svenseeberg (approved):

PR #77, opened 2023-07-05
Remove unnecessary permissions from models

### LINKED ISSUE #56: Revisit model permissions

charludo:
### Short description

Remove all custom model permissions (and some default once) since they were superfluous.

### Proposed changes

- remove all custom permissions
- remove a number of defualt permissions that are currently neither in use nor planned to be in use

### Side effects

- none I am aware of 

### Resolved issues

Fixes: #56


svenseeberg (approved):
This looks reasonable to me

timobrembeck (approved):

svenseeberg:
Oh damnit :scream_cat:

PR #80, opened 2023-07-07
Production preparation

svenseeberg:
### Short description
This is a summary PR of small changes to run this project on a server.

### Proposed changes
- Extract HOSTNAME from BASE_URL and add to ALLOWED_HOSTS
- Add MANIFEST.in file to include non-Python files in package

### Side effects
- N/A

### Resolved issues
- N/A

### DIFF
--- a/integreat_compass/core/settings.py
+++ b/integreat_compass/core/settings.py
@@ -12,6 +12,7 @@ https://docs.djangoproject.com/en/4.1/ref/settings/
 import os
 from distutils.util import strtobool
 from pathlib import Path
+from urllib.parse import urlparse
 
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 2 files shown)


integreat_compass/core/settings.py
```diff
@@ -31,7 +32,14 @@
 SECRET_KEY = os.environ.get("INTEGREAT_COMPASS_SECRET_KEY", "dummy" if DEBUG else "")
 
 BASE_URL = os.environ.get("INTEGREAT_COMPASS_BASE_URL", "http://localhost:8082")
-ALLOWED_HOSTS = [".localhost", "127.0.0.1", "[::1]"]
+HOSTNAME = urlparse(BASE_URL).hostname
+
+#: This is asecurity measure to prevent HTTP Host header attacks, which are possible even under many seemingly-safe
```
timobrembeck:
```suggestion
#: This is a security measure to prevent HTTP Host header attacks, which are possible even under many seemingly-safe
```

svenseeberg: Thanks

timobrembeck (approved):
:+1:

PR #82, opened 2023-07-12
Manage users via CRUD in production

### LINKED ISSUE #42: Manage User Accounts
### Motivation
We somehow need to manage access to the tool. Can we use the default CRUD back end? This will suffice for the beginning. Who is going to manage access in the long term?

### User Story
As a project admin I want to create users in the Django back end and assign a user role.

### Proposed Solution
- We can use the Django back end for the MVP.
- We need to have roles (board member, counselor, organization representative).
- The password reset e-mail notification should work to immediately trigger it for new user accounts.

### Alternatives
- None

svenseeberg:
### Short description
In #42 we decided to manage users via CRUD for the MVP. This PR always registers the user and permissions model in the CRUD back end.

### Proposed changes
- Register user model when DEBUG is False. 

### Side effects
- N/A

### Resolved issues
Fixes: #42

### DIFF
--- a/integreat_compass/cms/admin.py
+++ b/integreat_compass/cms/admin.py
@@ -4,6 +4,7 @@ Debug lists and forms for all models
 from django.apps import apps
 from django.conf import settings
 from django.contrib import admin
+from django.contrib.auth import get_user_model
 
 if settings.DEBUG:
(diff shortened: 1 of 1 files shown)


charludo (approved):
rubberstamped

PR #86, opened 2023-07-25
Replace numeric ids by natural keys in test data fixture

### LINKED ISSUE #83: Tests fail after updating permissions
### Describe the Bug
After merging #77 the tests on our `develop` branch fail because of missing permissions.

### Steps to Reproduce
https://app.circleci.com/pipelines/github/digitalfabrik/integreat-compass/385/workflows/1dacc0ab-4f68-4f79-8f7b-2d70a8ccb309/jobs/3160

### Expected Behavior
Tests should not fail :rainbow: 

### Actual Behavior
Tests fail :scream_cat: 

### Additional Information

<details>
  <summary>Traceback</summary>

  ```
=========================== short test summary info ============================
FAILED tests/cms/views/test_view_status_code.py::test_view_status_code_1[offer_provider@integreat.compass-cms:protected:new_offer-kwargs2-post_data2-roles2] - AssertionError: GET view cms:protected:new_offer returned status code 403 instead of 200 for role offer_provider@integreat.compass
FAILED tests/cms/views/test_view_status_code.py::test_view_status_code_1[offer_provider@integreat.compass-cms:protected:new_offer-kwargs3-post_data3-roles3] - AssertionError: POST view cms:protected:new_offer returned status code 403 instead of 302 for role offer_provider@integreat.compass
FAILED tests/cms/views/test_view_status_code.py::test_view_status_code_1[offer_provider@integreat.compass-cms:protected:edit_offer-kwargs4-post_data4-roles4] - AssertionError: POST view cms:protected:edit_offer returned status code 403 instead of 302 for role offer_provider@integreat.compass
========================= 3 failed, 22 passed in 2.52s =========================
  ```
</details>

charludo:
### Short description

As per @timoludwig's comment, replace the numeric ids of permissions by their natural keys in the `test_data.json` fixture.

### Proposed changes

- use natural keys
- remove the fallback image now that #71 is merged

### Side effects

/
 

### Resolved issues

Fixes: #83


timobrembeck:
Is it even required to duplicate the permission info here, or could we just remove it, since the group and permission info is already contained in the permissions.json fixture?

charludo:
> Is it even required to duplicate the permission info here, or could we just remove it, since the group and permission info is already contained in the permissions.json fixture?

Removing the permissions inside `test-data.json` breaks the tests in the same way as before, so I guess not, sadly.

timobrembeck:
But wouldn't it make sense to just load the permissions fixture in the tests as well?

charludo:
> But wouldn't it make sense to just load the permissions fixture in the tests as well?

...yes. Super sorry, I did not realize we were not doing so 🤦🏼‍♀️

integreat_compass/cms/fixtures/test_data.json
```diff
@@ -248,24 +248,21 @@
     "model": "auth.group",
     "pk": 1,
     "fields": {
-      "name": "BOARD_MEMBER",
-      "permissions": [38, 37]
+      "name": "BOARD_MEMBER"
```
timobrembeck: The permissions fixture also contains the groups itself, right? Maybe we could remove it here completely, so we don't get the wrong permissions when the fixtures are imported in the wrong order?

charludo: At least in the tests, it does not seem to make a difference which fixture is loaded first. Removing them completely from the test data makes all tests error out (`Problem installing fixtures: insert or update on t
able "cms_user_groups" violates foreign key constraint`) in the current loading order, but not when swapping them around.

timobrembeck: I think the correct logic would be to load the groups & permissions first, and then the rest of the test data (without any group information) afterwards.

charludo: Alright

timobrembeck (approved):
:+1:

tests/conftest.py
```diff
@@ -43,8 +44,8 @@ def load_test_data_transactional(transactional_db, django_db_blocker):
     :type django_db_blocker: :fixture:`django_db_blocker`
     """
     with django_db_blocker.unblock():
-        call_command("loaddata", "integreat_compass/cms/fixtures/test_data.json")
-        call_command("loaddata", "integreat_compass/cms/fixtures/permissions.json")
+        call_command("loaddata", "integreat_cms/cms/fixtures/permissions.json")
+        call_command("loaddata", "integreat_cms/cms/fixtures/test_data.json")
```
```suggestion
        call_command("loaddata", "integreat_compass/cms/fixtures/permissions.json")
        call_command("loaddata", "integreat_compass/cms/fixtures/test_data.json")
```

timobrembeck:
```suggestion
        call_command("loaddata", "permissions")
        call_command("loaddata", "test_data")
```
This should also work

timobrembeck (changes requested):
The conflict resolution did not really work, the group information is now contained in the test data again:

https://github.com/digitalfabrik/integreat-compass/blob/38044d05be90f71d895b4d4b026ab5fafc5eade1/integreat_compass/cms/fixtures/test_data.json#L373-L396

charludo:
> The conflict resolution did not really work, the group information is now contained in the test data again:
> 
> https://github.com/digitalfabrik/integreat-compass/blob/38044d05be90f71d895b4d4b026ab5fafc5eade1/integreat_compass/cms/fixtures/test_data.json#L373-L396

🤦🏼‍♀️ Sorry for this tedious review and wasting your time here.... Today's not my day 😅 

Also I gues we could (should?) eventually do the same thing for group ids/natural keys, but I think a custom group manager would be needed (?) and I'd rather push that into the future.

tests/conftest.py
timobrembeck: This change got reverted again with the latest push

timobrembeck (changes requested):
> Also I gues we could (should?) eventually do the same thing for group ids/natural keys, but I think a custom group manager would be needed (?) and I'd rather push that into the future.

I don't think that's necessary, since the groups are not auto-generated by the auth app (in comparison to the permission objects), the ids won't change as long as the order in the permissions fixture is not changed.

timobrembeck (approved):
Thanks! :+1:

PR #90, opened 2023-08-01
Add password reset for users

### LINKED ISSUE #78: Add password reset for users
### Motivation

Users should be able to reset their own passwords, instead of having to ask an admin.

### User Story

As a user I ~want to constantly~ will eventually forget my password. I want to be able to quickly reset it.

### Proposed Solution

Add a "forgot password" link to the login form. Have the user enter their email, then mail them a reset link, upon clicking which they have to set a new password.

### Alternatives

Not really any

charludo:
### Short description

Allow users to reset their passwords if they forgot it.

### Proposed changes

- Add hint to the login form
- add PW reset flow

### Side effects

- none I could find

### Resolved issues

Fixes: #78

### DIFF
--- a/integreat_compass/cms/utils/email_utils.py
+++ b/integreat_compass/cms/utils/email_utils.py
@@ -7,7 +7,10 @@ from django.utils.encoding import force_bytes
 from django.utils.http import urlsafe_base64_encode
 from django.utils.translation import gettext as _
 
-from .account_activation_token_generator import account_activation_token_generator
+from .token_generator import (
+    account_activation_token_generator,
+    password_reset_token_generator,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -59,12 +62,11 @@ def send_activation_mail(user):
     :type user: ~integreat_compass.cms.models.user.User
     """
     context = {"user": user}
-    subject = "Integreat Compass | Confirm your account"
+    subject = _("Integreat Compass | Activate your account")
 
     token = account_activation_token_generator.make_token(user)
     uid = urlsafe_base64_encode(force_bytes(user.pk))
 
-    subject += _("Activate your account")
     debug_mail_type = _("activation mail")
     context.update({"uid": uid, "token": token})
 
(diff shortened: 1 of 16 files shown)


integreat_compass/locale/de/LC_MESSAGES/django.po
```diff
@@ -672,7 +719,20 @@ msgstr "Aktivieren Sie Ihren Account"
...
+msgid "Integreat Compass | Your requested password reset"
+msgstr "Integreat Compass | Ihre angeforderte Passwortrücksetzung"
+
+#: cms/utils/email_utils.py
+msgid "Reset your password"
+msgstr "Passwort zurücksetzen"
+
+#: cms/utils/email_utils.py
+msgid "password reset mail"
+msgstr "Passwortrücksetzungsemail"
```
svenseeberg: While grammatically correct, I think this word is a bit too long and therefore hard to read. IMHO we should either insert a dash or split it: `Passwort-Zurücksetzen-Mail` or `Mail zum Passwort zurücksetzen`.

integreat_compass/locale/de/LC_MESSAGES/django.po
```diff
@@ -699,6 +760,30 @@ msgid ""
 msgstr ""
 "Bitte kontaktieren Sie einen Administrator um einen neuen Link anzufordern."
 
+#: cms/views/authentication/password_reset_confirm_view.py
+msgid "This password reset link is invalid."
+msgstr "Der Passwortrücksetzungslink ist ungültig."
```
svenseeberg: Same here. Passwortrücksetzungslink could be `Passwort-Zurücksetzen-Link`.

integreat_compass/cms/utils/email_utils.py
```diff
@@ -81,3 +84,35 @@ def send_activation_mail(user):
...
+    :param user: the user
+    :type user: ~integreat_compass.cms.models.user.User
+    """
+    context = {"user": user}
+    subject = _("Integreat Compass | Your requested password reset")
+
+    token = password_reset_token_generator.make_token(user)
+    uid = urlsafe_base64_encode(force_bytes(user.pk))
+
+    subject += _("Reset your password")
```
svenseeberg: AFAICT there is no space when concatenating the strings `"Integreat Compass | Your requested password reset"` and `"Reset your password"`. Do I miss something?

svenseeberg (changes requested):
Thanks! The PR looks very clean to me. Variable, class & method names are very good and easy to understand. Overall structure also looks sane to me. I only have some nitpicky comments on strings/translations ;-)

integreat_compass/cms/utils/email_utils.py
charludo: Uh, yes, you're right. I'm actually not sure why this is there, so I removed it. This part was copied from Integreat, I think it might worth checking out if the same thing is happening over there?

svenseeberg (approved):

Issue #74, opened 2023-06-28
Rework design to match the Integreat CI


JoeyStk:
### Motivation
Since the project has been rebranded from Integrationskompass to Integreat-Kompass we need to make some adjustments to the UI to match the CI of Integreat.

### User Story
As a project owner I want the user to have a subconscious connection to Integreat so that they also find other related products/assistance.

### Proposed Solution
Follow the new guidelines defined by the UI/UX team in Figma

### Alternatives
None

### Additional Context
There is already some work on this in progress, but as for now this issue is still blocked by UI/UX

### Screenshots
Screenshot for: All offers
![grafik](https://github.com/digitalfabrik/integreat-compass/assets/72705147/aa60e682-de54-4f0f-999a-884db61b76fa)

(More screenshots coming as work on this progresses)

PR #98, opened 2023-08-14
Add reporting functionality

### LINKED ISSUE #88: Add reporting functionality
### Motivation

Users should be able to report offers.

### User Story

As a integration specialist or non-logged in user, I want to be able to report offers.

### Proposed Solution

Implement reporting functionality from the details overlay of offers. TBD: should a reason be providable / need to be provided?

### Screenshots

image

JoeyStk:
### Short description
This PR adds the option to report offers

### Proposed changes

- Add option to report on offer detail that directs to a reporting form
- In reporting form fill out why the offer should be reported

### Side effects

- During the initial setup / ER modelling we said that for now it's sufficient to have a textarea to comment. I can imagine that in the future we might do a bit more fine tuning but I think for the MVP this should be fine 
- Initially I wanted to redirect back to the index page and show a message for the form there. However with the sidebar the message wasn't properly showing. My solution was now to stay on the form page but with add the option to go back to the index page via an additional mini navigation
- All of this is still a bit ugly, but since we're going to change the UI with #74 anyway I thought it was okay for now

### Resolved issues

Fixes: #88

### DIFF
--- a/integreat_compass/cms/forms/offers/report_form.py
+++ b/integreat_compass/cms/forms/offers/report_form.py
@@ -1,0 +1,16 @@
+from ...models import Report
+from ..custom_model_form import CustomModelForm
+
+
+class ReportForm(CustomModelForm):
+    """
+    Form for reporting an offer
+    """
+
+    class Meta:
+        """
+        This class contains the meta information
+        """
+
+        model = Report
+        fields = ["comment"]
--- a/integreat_compass/cms/templates/offers/offer_report.html
+++ b/integreat_compass/cms/templates/offers/offer_report.html
@@ -1,0 +1,27 @@
+{% extends "_base.html" %}
+{% load i18n %}
+{% load static %}
+{% load widget_tweaks %}
+{% load render_bundle from webpack_loader %}
+{% block title %}
+    {% blocktranslate trimmed with offer_title=selected_offer.public_version.title %}
+        Report offer "{{ offer_title }}"
+    {% endblocktranslate %}
+{% endblock title %}
+{% block content %}
+    <div class="w-full flex flex-wrap justify-between mb-6 gap-2">
+        <h1 class="heading">
+            {% blocktranslate trimmed with offer_title=selected_offer.public_version.title %}
+                Report offer "{{ offer_title }}"
+            {% endblocktranslate %}
+        </h1>
+    </div>
+    <div class="grid xl:grid-cols-2 gap-x-24">
+        <form method="post">
+            {% csrf_token %}
+            <label for="{{ form.comment.id_for_label }}">{% translate "Reason for reporting" %}</label>
+            {{ form.comment }}
+            <button type="submit" class="btn mt-2">{% translate "Report offer" %}</button>
+        </form>
+    </div>
+{% endblock content %}
--- a/integreat_compass/cms/views/offers/offer_report_view.py
+++ b/integreat_compass/cms/views/offers/offer_report_view.py
@@ -1,0 +1,75 @@
+import logging
+
+from django.contrib import messages
+from django.shortcuts import get_object_or_404, redirect, render
+from django.utils.translation import gettext as _
+from django.views.generic import TemplateView
+
+from ...forms import ReportForm
+from ...models import Offer, OfferVersion
+
+logger = logging.getLogger(__name__)
+
+
+class OfferReportView(TemplateView):
+    """
+    View to report an offer
+    """
+
+    template_name = "offers/offer_report.html"
+
+    def get_context_data(self, **kwargs):
+        r"""
+        Add context data for the offer report view
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: The context used in the offer report view
+        :rtype: dict
+        """
+        return super().get_context_data(**kwargs) | {
+            "selected_offer": get_object_or_404(Offer, id=kwargs.get("pk")),
+            "form": ReportForm(),
+        }
+
+    def post(self, request, *args, **kwargs):
+        r"""
+        This functions handles the input from the report function
+
+        :param request: The current request
+        :type request: ~django.http.HttpRequest
+
+        :param \*args: The supplied arguments
+        :type \*args: list
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: Redirect to the index view
+        :rtype: ~django.http.HttpResponseRedirect
+        """
+        offer = get_object_or_404(Offer, id=kwargs.get("pk"))
+        form = ReportForm(data=request.POST)
+
+        if not form.is_valid():
+            form.add_error_messages(request)
+            return render(
+                request,
+                self.template_name,
+                {"form": form, **self.get_context_data(**kwargs)},
+            )
+
+        form.instance.offer_version = OfferVersion.objects.get(
+            id=offer.public_version.id
+        )
+        form.save()
+
+        messages.success(
+            request,
+            _(
+                'Your report for offer "{}" has successfully been submitted and will be reviewed shortly.'
+            ).format(offer.public_version.title),
+        )
+
+        return redirect("cms:public:index")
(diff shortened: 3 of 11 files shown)


integreat_compass/cms/forms/offers/report_form.py
```diff
@@ -0,0 +1,18 @@
+from django import forms
+
+from ...models import Report
+
+
+class ReportForm(forms.ModelForm):
+    """
+    Form for reporting an offer
+    """
```
charludo:
```suggestion
from django import forms

from ...models import Report
from ..custom_model_form import CustomModelForm


class ReportForm(CustomModelForm):
    """
    Form for reporting an offer
    """
```
No reason not to use then fancy custom class we already use everywhere else 😄

integreat_compass/cms/templates/offers/offer_report.html
```diff
@@ -0,0 +1,32 @@
...
+        <h1 class="heading">
+            {% blocktranslate trimmed with offer_title=selected_offer.public_version.title %}
+                Report offer "{{ offer_title }}"
+            {% endblocktranslate %}
+        </h1>
+    </div>
+    <div class="flex items-center">
+        <i class="text-primary" icon-name="chevron-left"></i>
+        <a class="text-primary hover:cursor-pointer font-bold hover:underline"
+           href="{% url 'cms:public:index' %}">{% translate "Go back to all offers" %}</a>
```
charludo: IMO this link has to redirect to the filtered view the user was seeing previously. Currently all filters are lost.

The proper way would probably be to pass the previous URL along when hitting the report button. An easier (albeit a bit "dirtier") way could be to just use JS and do `history.back()`, though if you find the time, I'd prefer the former method 😅

integreat_compass/cms/views/offers/offer_report_view.py
```diff
@@ -0,0 +1,87 @@
...
+    def _get_form():
+        """
+        This function gets the form Report
+
+        :return: the report form
+        :rtype: dict
+        """
+        form = ReportForm()
+        return form
+
```
charludo: I think this was copied and modified from the `OfferFormView`, correct? There we used this method because we had to initialize so many forms both in GET and POST methods. It's not necessary here (thankfully)
```suggestion
```

integreat_compass/cms/views/offers/offer_report_view.py
```diff
@@ -0,0 +1,87 @@
...
+        :param kwargs: The supplied keyword arguments
+        :type kwargs: dict
+
+        :return: The context of all
+        :rtype: dict
+        """
+        offer = Offer.objects.get(id=kwargs.get("pk"))
+        context = {"selected_offer": offer, "form": self._get_form()}
+        return render(request, self.template_name, context)
```
charludo:
```suggestion
    def get_context_data(self, **kwargs):
        r"""
        Add context data for the offer report view

        :param \**kwargs: The supplied keyword arguments
        :type \**kwargs: dict

        :return: The context used in the offer report view
        :rtype: dict
        """
        return super().get_context_data(**kwargs) | {
            "selected_offer": Offer.objects.get(id=kwargs.get("pk")),
            "form": ReportForm(),
        }
```
`get_context_data` should be enough

integreat_compass/cms/views/offers/offer_report_view.py
```diff
@@ -0,0 +1,87 @@
...
+            form.save()
+
+        messages.success(
+            request,
+            _(
+                'Your report for offer "{}" has successfully been submitted and will be reviewed shortly.'
+            ).format(offer.public_version.title),
+        )
+
+        return redirect("cms:public:index")
```
charludo:
```suggestion
    def post(self, request, *args, **kwargs):
        r"""
        This functions handles the input from the report function

        :param request: The current request
        :type request: ~django.http.HttpRequest

        :param \*args: The supplied arguments
        :type \*args: list

        :param \**kwargs: The supplied keyword arguments
        :type \**kwargs: dict

        :return: Redirect to the index view
        :rtype: ~django.http.HttpResponseRedirect
        """
        offer = Offer.objects.get(id=kwargs.get("pk"))
        form = ReportForm()

        if not form.is_valid():
            form.add_error_messages(request)
            return render(
                request,
                self.template_name,
                {"form": form, **self.get_context_data(**kwargs)},
            )

        form.instance.offer_version = OfferVersion.objects.get(
            id=offer.public_version.id
        )
        form.save()

        messages.success(
            request,
            _(
                'Your report for offer "{}" has successfully been submitted and will be reviewed shortly.'
            ).format(offer.public_version.title),
        )

        return redirect("cms:public:index")
```

Again, I think the `for form in [*form.values()]:` is a remnant of `OfferFormView`, correct? Since we only have one form here, there's no need to loop. 

I'd also suggest rendering the same view again with the error messages instead of redirecting to index (and showing the success message!) even when an error occurred.

Then you also don't have to check the form validity again.

charludo (changes requested):
Thank you! I really like the way the reporting flow looks and feels now, the separate page was a good idea 👍🏼 

Some suggestions, mostly for code readability:

integreat_compass/cms/views/offers/offer_report_view.py
charludo: I just realized you also need to check if the PK exists, currently a DoesNotExist exception is thrown. `.filter(id=kwargs.get("pk")).first()` could be a drop-in replacement for `.get(...)`, I don't know if we have to handle users artificially changing the PK in their address bar beyond that...

integreat_compass/cms/views/offers/offer_report_view.py
charludo: Same here with the handling of non-existant offer PKs.

integreat_compass/cms/templates/offers/offer_report.html
JoeyStk: I solved it via request.META.get('HTTP_REFERER'). I hope that is not worrisome

integreat_compass/cms/views/offers/offer_report_view.py
JoeyStk: This case I handled through an try-except. I hope this is sufficient

integreat_compass/cms/views/offers/offer_report_view.py
JoeyStk: I changed it to get_object_or_404. I hope that is sufficient

integreat_compass/cms/views/offers/offer_report_view.py
JoeyStk: With the get_object_or_404 the try-except might be redundant, but I'm not entirely sure, so for now I kept it

JoeyStk:
Changes are applied. Could you please review again? :smile:

integreat_compass/cms/templates/offers/offer_report.html
charludo: Good idea! Only issue I am seeing right now is that after successfully submitting the report, that link now points to the reporting URL.

integreat_compass/cms/views/offers/offer_report_view.py
charludo: `get_object_or_404`is a great idea. I don't think the try-except is necessary now, so I'd say it can be removed + we no longer need an explicit else case

integreat_compass/cms/templates/offers/offer_report.html
JoeyStk: Ah damn, yes. I will have a look at it

JoeyStk: This is now also resolved as discussed in the weekly meeting. If the submitting is successful the user gets redirected to the index page and gets the success message there. The issue for the overlapping filter bar is already created, see [here](https://github.com/digitalfabrik/integreat-compass/issues/100).

JoeyStk:
Changes are applied again. Please review this PR again :)
The latest commit also introduces the image height on the index page and in the detail overlay :)

charludo (approved):
Awesome, thank you! 🎉

PR #99, opened 2023-08-16
Fix fallback title image not being shown

charludo:
### Short description

If an offer used the default fallback image, it was not shown in the list view or detail view.

### Proposed changes

- use the `title_image_url` property introduced in #71 instead of `title_image.url`

### Side effects

- none

### Resolved issues

Fixes: n/a


JoeyStk (approved):
Somehow I didn't realize this change when My Offers was merged. Thank you for the fix! :rocket:

PR #102, opened 2023-08-18
Add location and distance filter

### LINKED ISSUE #97: Add distance/location filter
### Motivation
Users should be able to filter offers by geographical location.

### User Story
As a job center councilor I want to filter offers by location so that the offers are physically reachable for my clients.

### Proposed Solution
- Add an input field that limits the distance of shown offers by distance to my location (default 25km).
- Add a Location setting to the offer filters (which is prefilled by the My Location setting from the profile)
- Use Nominatim API to resolve an entered address to coordinates
- Optional (prio low): Add a "My Location" setting in the profile
- Optional (prio low): use GPS to fill current location field

### Alternatives
- N/A

### Additional Context
- We could use [GeoPy](https://geopy.readthedocs.io/en/stable/)
- #39 

### Screenshots
image

charludo:
### Short description

Add location and distance filter.

Note this does NOT pre-fill the address search field with anything, since as mentioned in the issue, we have none of the views/functionalities required for that (yet?)

### Proposed changes

- add address search field, map and radius toggles to the filter bar
- when an address and radius are selected, filter offers accordingly

### Side effects

- had to refactor some of the map code

### Resolved issues

Fixes: #97

### DIFF
--- a/integreat_compass/cms/templates/index.html
+++ b/integreat_compass/cms/templates/index.html
@@ -8,7 +8,7 @@
 {% endblock container %}
 {% block content %}
     <div class="flex">
-        <div class="pl-12 hidden lg:block fixed top-24 bottom-0 bg-white lg:w-3/12 py-8 pr-8 shadow-[10px_0_5px_-2px_rgba(0,0,0,0.1)]">
+        <div class="pl-12 hidden lg:block fixed top-24 bottom-0 bg-white lg:w-3/12 py-8 pr-8 shadow-[10px_0_5px_-2px_rgba(0,0,0,0.1)] overflow-y-scroll">
             <form id="filter-form" action="{% url 'cms:public:index' %}" method="get">
                 <h2 class="text-2xl font-semibold">{% translate "Filter" %}</h2>
                 <p class="text-gray-500">{% translate "Filter all offers" %}</p>
@@ -16,6 +16,7 @@
                     <input id="search"
                            placeholder="{% translate "Search" %}"
                            name="search"
+                           value="{% if filtered_search %}{{ filtered_search }}{% endif %}"
                            type="text"
                            class="filter-element block w-full px-1 border border-gray-400 rounded-md">
                 </div>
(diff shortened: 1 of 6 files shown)


integreat_compass/cms/templates/index.html
```diff
@@ -79,6 +80,46 @@ <h3 class="text-gray-500 mt-10">{% translate "Categories" %}</h3>
...
+                                   step="1e-7"
+                                   id="id_location-lat">
+                            <input type="number"
+                                   class="filter-element"
+                                   name="long"
+                                   step="1e-7"
+                                   id="id_location-long">
+                        </div>
+                        <ul class="mt-2 items-center w-full border border-gray-500 rounded flex shadow-inner">
+                            {% for distance in offer_distance_steps %}
```
JoeyStk, line 118: This section also overflows for me on smaller laptop / tablet screens. 
image

This gets even worse if we agree to add the overflow-y-scroll as mentioned above, because I rather have an overflowing element than have it vanish on smaller screens. Unfortunately I wasn't able to come up with a quick yet efficient solution for it.

JoeyStk (changes requested):
Once again, amazing work. Thank you very much! I only have a few thought for some details:

- I know that for now we are not focusing on smaller devices, but the filter is causing issues also on smaller laptop screens (1024px - ~1600px) in so far that in Firefox there is no default vertical scrollbar. With that it wasn't possible for me to click on the map. My idea would have been to add 'overflow-y-scroll' to [this line]( https://github.com/digitalfabrik/integreat-compass/blob/develop/integreat_compass/cms/templates/index.html#L11). This fixes the issue, however it creates a new one as described in my comment below. Maybe you have a different and better idea?

- Lastly I would have expected to be able to reset the filter in some form or another. I wasn't able to reset the filter in a different way than erasing the get parameters from the url bar. I don't think this point is super important, but if it can be done quickly I think it would be nice, that if the input text field for the address is empty and I click outside of the textfield, that the filter gets reset. However this is surely debatable. What do you think?

And lastly, I couldn't find any side effects to the refactoring of the code for the map :)

JoeyStk (approved):
Really cool improvements! Thank you very much :)

PR #105, opened 2023-08-21
Add reported and declined offers list

### LINKED ISSUE #89: Show blocked and declined offers to board members
### Motivation

In the applications view available to board members, three tabs exist (applications, reports, declined). Only the first one will be populated in #61.

### User Story

As a board member, I want to be able to see which offers have been reported by users or rejected by myself in the past in case I change my mind.

### Proposed Solution

Implement the views for these tabs

### Additional Context

Blocked by:
- [x] #38  
- [x] #88

### Screenshots

No mockup provided in Figma, but should be the same as the main applications tab:
image

JoeyStk:
### Short description
This PR adds the list of reported and declined offers to the applications navigation. 

### Proposed changes

- Add view for reported offers
- Add view for declined offers

### Side effects

- @svenseeberg and I discussed that it might make sense to show the report also to the creator of the offer in question. Since I have already worked on it but not yet completed it, I thought I'd open up [another branch for it](https://github.com/digitalfabrik/integreat-compass/tree/feature/add_report_list_with_details), if at one point we decide to revise it.

### Resolved issues

Fixes: #89

### DIFF
--- a/integreat_compass/cms/templates/offers/offer_report.html
+++ b/integreat_compass/cms/templates/offers/offer_report.html
@@ -15,6 +15,9 @@
                 Report offer "{{ offer_title }}"
             {% endblocktranslate %}
         </h1>
+        <p>
+            {% translate "Please note: You will stay anonymous. The report will be seen by both the offer provider and board members." %}
+        </p>
     </div>
     <div class="grid xl:grid-cols-2 gap-x-24">
         <form method="post">
--- a/integreat_compass/cms/views/interactions/report_list_view.py
+++ b/integreat_compass/cms/views/interactions/report_list_view.py
@@ -1,0 +1,64 @@
+from django.contrib import messages
+from django.shortcuts import redirect
+from django.utils.decorators import method_decorator
+from django.utils.translation import gettext as _
+from django.views.generic import TemplateView
+
+from ...decorators import permission_required
+from ...models import OfferVersion, Report
+
+
+@method_decorator(permission_required("cms.add_vote"), name="dispatch")
+class ReportListView(TemplateView):
+    """
+    View for the report list
+    """
+
+    template_name = "interactions/report_list.html"
+
+    def get_context_data(self, **kwargs):
+        r"""
+        Context for the report list
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        return super().get_context_data(**kwargs) | {
+            "reports": [
+                report
+                for report in Report.objects.filter()
+                if report.offer_version == report.offer_version.offer.public_version
+            ]
+        }
+
+    def post(self, request, *args, **kwargs):
+        r"""
+        Handle form to declare a reported offer as declined
+
+        :param request: The current requestHttpResponseRedirect
+        :type request: ~django.http.HttpRequest
+
+        :param \*args: The supplied arguments
+        :type \*args: list
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.HttpResponseRedirect
+
+        """
+        offer_version_id = request.POST.get("offer_version_id")
+        if request.POST.get("move-to-declined"):
+            OfferVersion.objects.filter(id=offer_version_id).update(state="False")
+
+        messages.success(
+            request,
+            _('You rejected the reported offer "{}"').format(
+                OfferVersion.objects.get(id=offer_version_id).title
+            ),
+        )
+        return redirect("cms:protected:reports")
--- a/integreat_compass/cms/fixtures/test_data.json
+++ b/integreat_compass/cms/fixtures/test_data.json
@@ -86,6 +86,20 @@
     }
   },
   {
+    "model": "cms.offerversion",
+    "pk": 6,
+    "fields": {
+      "offer": 5,
+      "created_at": "2023-08-17T11:33:22.346Z",
+      "title": "Krisenchat: Psychologischer Support per Textnachricht",
+      "title_image": "",
+      "description": "Psychologische Unterschützung per Textnachricht rund um die Uhr und in Echtzeit",
+      "is_free": true,
+      "language": 2,
+      "state": false
+    }
+  },
+  {
     "model": "cms.user",
     "pk": 1,
     "fields": {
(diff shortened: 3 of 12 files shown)


integreat_compass/cms/views/components/decline_list_view.py
```diff
@@ -0,0 +1,25 @@
...
+        Context for the list of declined offers
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        return super().get_context_data(**kwargs) | {
+            "declined_offers": OfferVersion.objects.filter(state="False")
```
JoeyStk: This query is not exactly correct yet. It should be offers who latest version has been rejected not all rejected offers

integreat_compass/cms/views/components/report_list_view.py
```diff
@@ -0,0 +1,58 @@
...
+        r"""
+        Context for the report list
+
+        :param \**kwargs: The supplied keyword arguments
+        :type \**kwargs: dict
+
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        return super().get_context_data(**kwargs) | {"reports": Report.objects.all()}
```
JoeyStk: Same goes here: This query needs to be a bit different. I should be all offers who's latest version has been reported and not been rejected yet

integreat_compass/cms/templates/components/decline_list.html
```diff
@@ -0,0 +1,28 @@
...
+                <div class=" my-4">
+                    {% for tag in offer.offer.tags.all %}
+                        <span class="bg-red-500 text-white rounded-full py-1 px-2">{{ tag }}</span>
+                    {% endfor %}
+                </div>
+                <p class="text-gray-700">{% translate "Created at:" %} {{ offer.created_at|date:"d. M Y" }}</p>
+            </div>
+            <div class="basis-1/3 flex items-center"></div>
+        </div>
+    {% endfor %}
```
charludo:
```suggestion
    {% empty %}
        {% translate "No offers have been declined by the board." %}
    {% endfor %}
```

(or something along the lines)

integreat_compass/cms/templates/components/report_list.html
```diff
@@ -0,0 +1,46 @@
...
+                        <button class="bg-blue-500 hover:bg-blue-700 rounded-full" type="button">{% translate "Details" %}</button>
+                        <button class="bg-red-500 hover:bg-red-700 rounded-full"
+                                type="submit"
+                                name="move-to-declined"
+                                value="True">{% translate "Reject version" %}</button>
+                    </div>
+                </form>
+            </div>
+        </div>
+    {% endfor %}
```
charludo:
```suggestion
    {% empty %}
        {% translate "There are no unhandled reports." %}
    {% endfor %}
```

(or something along the lines)

integreat_compass/cms/templates/offers/offer_report.html
```diff
@@ -15,6 +15,7 @@ <h1 class="heading">
                 Report offer "{{ offer_title }}"
             {% endblocktranslate %}
         </h1>
+        <p>{% translate "Please note: This report can be seen by both the offer provider and board members." %}</p>
```
charludo:
```suggestion
        <p>{% translate "Please note: You will stay anonymous. The report will be seen by both the offer provider and board members." %}</p>
```

(or something along the lines - I think the anonymous part is important because otherwise it could be misinterpreted as "better watch your tongue, we'll show this to lots of people!" 😄)

integreat_compass/cms/views/components/decline_list_view.py
```diff
@@ -0,0 +1,25 @@
...
+        }
```
charludo:
```suggestion
        return super().get_context_data(**kwargs) | {
            "declined_offers": OfferVersion.objects.filter(state="False").distinct(
                "offer"
            )
        }
```

I think in the long run (not the MVP 😄) we should show the reports in the voting form if the previous public version received reports as well. Otherwise Providers might add an offer, it gets approved, it gets reported, Provider changes some small thing not relating to the report, offer gets approved again because the Board is not aware of the previous report.

An edit history might even be a good idea 😄

integreat_compass/cms/views/components/report_list_view.py
charludo: I think this should work:

```suggestion
        return super().get_context_data(**kwargs) | {
            "reports": [
                report
                for report in Report.objects.filter()
                if report.offer_version
                == report.offer_version.offer.public_version
                == report.offer_version.offer.latest_version
            ]
        }
```
Only public versions can be reported, so initially public_version and report.offer_version are the same. Now if either the Offer Provider updates the offer (latest_version changes) or the board reacts to the report and rejects the offer (public_version changes) the offer disappears from the list.

Edge case: the Offer Provider has updated their offer prior to the report being made, but this change has not been approved yet. Then noone ever sees the report.

Maybe the equality check for latest_version should be removed? This would fix the edge case, and it is probably OK if a report stays visible until a new public version has been approved/rejected? Hm.

Edit: After rejecting, board members should no longer be shown the report. This runs into the same issue as the "ignore" thing though.

charludo (changes requested):
Thank you! This looks great.

Apart from the suggestions, I do not really like the `component/` subdirs - they are very nondescript. Could you either put the new files into `interactions/` (where they kind of fit.... kind of) or in a subdir of interactions, or come up with a more descript name? 😄 

Another thing that would maybe be good to add is an "ignore" button to the reports. Unjustified or spammy reports should not stay in the list forever.

Edit: that "ignore" would have to be user-specific though. Hm. If you have a good idea how to implement this without much effort (I do not), please go ahead! Otherwise, could you add a backlog ticket?

JoeyStk:
Thank you for your feedback. All issues should be addressed now. 

- I moved the files to interaction. I think it makes the most sense if they are in the same directory as their respective model is
- I opened an issue for the "ignore" button

integreat_compass/cms/views/interactions/report_list_view.py
```diff
@@ -0,0 +1,66 @@
...
+        :return: The rendered template response
+        :rtype: ~django.template.response.TemplateResponse
+        """
+        return super().get_context_data(**kwargs) | {
+            "reports": [
+                report
+                for report in Report.objects.filter()
+                if report.offer_version
+                == report.offer_version.offer.public_version
+                == report.offer_version.offer.latest_version
```
charludo:
```suggestion
```
Sorry to reopen this, I think this line needs to go, my bad. IMO it's OK for the report to stay until a new version has been approved or the current one rejected.

JoeyStk: Ah okay :+1:

JoeyStk: Is applied now! Thank you for all your help!!

charludo (approved):
Thank you!!

PR #106, opened 2023-08-21
Comment on offers

### LINKED ISSUE #41: Comment on offers
### Motivation
To ensure the quality of the content, users should be able to share experience about the quality of offers.

### User Story
As a job center counselor I want to ~~rate and~~ comment on the quality of an offer so that other job center counselors can assess the quality of an offer.

### Proposed Solution
- ~~Give counselors the option to rate offers between 1 and 5 stars and also enter free text comments.~~
- Both are visible to other counselors.
- Show the ratings and comments of other users.
- A "Save" button is needed.
- The average (stars scale) can be implemented either in JS in the front or back end. Front end: Transmit number to JS code and fill overlay. Back End: round to full or half numbers and use full or half full stars.

### Alternatives
- None

### Additional Context
- We have no wireframe for this functionality yet. Could be implemented in the details screen ( #40 ).
- Municipalities told us that rating is not possible due to legal constraints. We can just omit that functionality for now.  We can keep the rating in the database model for now, as we may implement this function after the MVP.
- The "Comment" should be translated as "Vermerk" into German to avoid this "rating feeling".

charludo:
### Short description

Add ability to comment on offers, and show those comments below offers.

### Proposed changes

- add commenting field for users with the required permission
- use ajax to post comments
- show a hint when a comment was made for an outdated offer version
- **fix** a bug where the state of new offer versions was not reset to PENDING (most likely introduced in #70)

### Side effects

- none that I could find

### Resolved issues

Fixes: #41


integreat_compass/cms/templates/overlay/offer-detail-overlay.html
```diff
@@ -82,6 +82,81 @@ <h3 class="text-xl mt-4 font-semibold">{% translate "More details" %}</h3>
...
+                                       value="{{ offer.pk }}">
+                                <button type="submit" class="btn mt-2">{% translate "Add comment" %}</button>
+                                <div class="float-right">
+                                    <span class="comment-length">0</span>/{{ max_comment_length }}
+                                </div>
+                            </form>
+                        </div>
+                    {% endif %}
+                    <div class="shrink grow-0 {% if not offer.comments %}hidden{% endif %}">
+                        <div class="mb-4">
```
JoeyStk:
```suggestion
                        <div class="my-4">
```
This is quite picky, but I would personally prefer for the two columns to start on the same height :sweat_smile: but feel free to disagree :blush:

integreat_compass/cms/fixtures/test_data.json
JoeyStk, line 1: Nice!

integreat_compass/cms/templates/overlay/offer-detail-overlay.html
```diff
@@ -82,6 +82,81 @@ <h3 class="text-xl mt-4 font-semibold">{% translate "More details" %}</h3>
...
+                                                <a href="{{ document.file.url }}" target="_blank" class="font-bold">{{ document.name }}</a>
+                                                <p>{{ offer.public_version.created_at }}</p>
+                                            </div>
+                                        </li>
+                                    {% endfor %}
+                                </ul>
+                            </div>
+                        </div>
+                    {% endif %}
+                </div>
```
JoeyStk:
```suggestion
                </div>
         {% endif %}
```
Of course only, if you choose to follow my suggestion above

integreat_compass/cms/templates/overlay/offer-detail-overlay.html
```diff
@@ -82,6 +82,81 @@ <h3 class="text-xl mt-4 font-semibold">{% translate "More details" %}</h3>
                         </div>
                     </div>
                 </div>
+                <div class="md:flex justify-left gap-12">
```
JoeyStk, line 88:
```suggestion
{% if public_version %}
                <div class="md:flex justify-left gap-12">
```
Admittedly  this is only an issue for the root user, which we don't really need to pay attention to, but the root user can access this section also via "Application > Details" (The board members are filtered out via next permission check in the next line) and in this view this section doesn't fully work, because max_comment_length doesn't exist for this view. Again this is not really an issue for non-developer-user, but since there is an easy fix for it by excluding non public versions, I say we add it?

integreat_compass/cms/forms/__init__.py
```diff
@@ -4,10 +4,10 @@
 """
 from .authentication.password_reset_request_form import PasswordResetRequestForm
 from .authentication.registration_form import RegistrationForm
+from .interactions.report_form import ReportForm
```
JoeyStk, line 7: Thanks for cleaning up after me! :see_no_evil:

integreat_compass/cms/urls/protected.py
```diff
@@ -44,4 +44,8 @@
             ]
         ),
     ),
+    path(
+        "comment/",
+        include([path("add/", interactions.add_comment, name="add_comment")]),
```
JoeyStk, line 49: I noticed that I, as any user (not logged in, board member, integration specialist & offer provider) can access this url (and get a blank page). This is where I'm unsure - is this harmless?

integreat_compass/locale/de/LC_MESSAGES/django.po
```diff
@@ -803,6 +803,26 @@ msgstr "Gruppentyp:"
 msgid "Mode type:"
 msgstr "Modustype:"
 
+#: cms/templates/overlay/offer-detail-overlay.html
+msgid "Comment on this offer"
+msgstr "Vermerk zur Maßnahme hinzufügen"
```
JoeyStk, line 820:
```suggestion
msgstr "Kommentar zur Maßnahme hinzufügen"
```
It's surely not incorrect to use "Vermerk", but I personally and the design prefer "Kommentar". I think that in the end it's up to you :)

JoeyStk (changes requested):
Thank you very much for this PR! It's a cool feature!
As far as I know all of the details and permissions work as expected. I only have a few minor comments below :)

integreat_compass/cms/templates/overlay/offer-detail-overlay.html
charludo: Good catch & nice fix!!

integreat_compass/cms/forms/__init__.py
charludo: no worries 😂

integreat_compass/locale/de/LC_MESSAGES/django.po
charludo: While I do agree, apparently this has been requested explicitly (see the very end of the issue description)

integreat_compass/cms/urls/protected.py
charludo: ~~good catch! No, I think that should not be possible 😅 ~~

actually I think it's fine, the GET method is just not defined at all.

integreat_compass/locale/de/LC_MESSAGES/django.po
JoeyStk: Ah sorry, that is on me :see_no_evil:

JoeyStk (approved):
Thank you! :tada:

Issue #107, opened 2023-08-21
Include missing offer details in the offer details overlay


charludo:
### Motivation

All relevant offer details should be shown (that's why we are asking for them after all! 😄)

### User Story

As a user, I want to get as much information as possible on each offer.

### Proposed Solution

Add the following bits of information which we are already collecting, but not showing to users:
- offer language
- is the offer free?
- group type
- mode type
- organization website
- (optional) a map with a pin at the address; this would be nice-to-have (since seeing a location on a map is much more intuitive than reading an unknown address, then having to search for the address on maps yourself), however we could also move this to the backlog

### Additional Context

Not sure if we need UI/UX for this - it's mostly just additional details lines in the left column.

JoeyStk:
Also resolved by #108

PR #108, opened 2023-08-21
Add details overlay to applications and add more fields

### LINKED ISSUE #87: Make offer details accessible from applications view
### Motivation

Board members should be able to quickly look at offer details.

### User Story

As a board member, I want to have easy access to all details about an offer when deciding on how to vote.

### Proposed Solution

Once #61 and #75 are merged, largely copy the popup introduced in #75, minus the eventual reporting and commenting functionalities, as well as the ratings by integration specialists.

### Alternatives

Make something new

### Additional Context

Blocked by:
- [x] #75 

### Screenshots

image

JoeyStk:
### Short description
This PR adds the detail overlay for offers to the application section. For the cases where we would access the version of an offer I created another variable called offer_version, which for "All Offers" is offer.public_version and for "Applications" is offer.latest_version. With that it was possible to stay DRY :)

### Proposed changes

- Extend the detail offer view to the application process 
- Make it possible to close popup when clicking on backdrop
- Rename trigger to open details offer from "single-offer" to "open-details" to make it clearer what happens when you add this class to an element
- Add image to view list

### Side effects

- I think none
- The last part where I see some room for improvement would be the naming of the variables in the offer-details.ts. In some cases I still use layover but we agreed to use overlay instead. If wanted, I can rename them and add it to this PR

### Resolved issues

Fixes: #87,  #107

### DIFF
--- a/integreat_compass/static/src/js/offer-details.ts
+++ b/integreat_compass/static/src/js/offer-details.ts
@@ -16,6 +16,14 @@ export const handleOfferDetails = () => {
                 }
             });
         }
+        if (offerDetailLayover) {
+            // Close window by clicking on backdrop.
+            offerDetailLayover.addEventListener("click", (e) => {
+                if (e.target === offerDetailLayover) {
+                    offerDetailLayover.classList.toggle("hidden");
+                }
+            });
+        }
     }
 };
 
--- a/integreat_compass/static/src/js/offer-details.ts
+++ b/integreat_compass/static/src/js/offer-details.ts
@@ -1,5 +1,5 @@
 export const handleOfferDetails = () => {
-    const allOffers = document.getElementsByClassName("single-offer");
+    const allOffers = document.getElementsByClassName("open-details");
     for (const offer of allOffers) {
         const id = offer.id.replace("offer-", "");
         const offerDetailLayover = document.getElementById(`offer-detail-layover-${id}`);
--- a/integreat_compass/cms/templates/index.html
+++ b/integreat_compass/cms/templates/index.html
@@ -96,7 +96,7 @@
                     {% if offers %}
                         {% for offer in offers %}
                             <div id="offer-{{ offer.id }}"
-                                 class="single-offer bg-white flex shadow-lg hover:shadow-xl rounded-2xl hover:cursor-pointer">
+                                 class="open-details bg-white flex shadow-lg hover:shadow-xl rounded-2xl hover:cursor-pointer">
                                 <div class="w-4/12 flex items-center">
                                     <img class="rounded-2xl object-cover h-[300px]"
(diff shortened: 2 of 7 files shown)


charludo:
Thank you! 🎉 

> However the solution we thought of last week didn't work as expected and therefore I went back to the dirty version.

Could you elaborate a bit on the issue? 😅

JoeyStk:
> Could you elaborate a bit on the issue? 😅

After thinking about this again today and a bit more critically I found a way to reuse all of the code by simply creating a new variable called offer_version which in one cases is equal to offer.public_version and in the other to offer.latest_version. With this we can use the same template in both cases. It's embarrassing that I didn't think of it yesterday :see_no_evil:  I think the PR is now ready for review

integreat_compass/static/src/js/offer-details.ts
```diff
@@ -16,6 +16,14 @@ export const handleOfferDetails = () => {
...
             });
         }
+        if (offerDetailLayover) {
+            // Close window by clicking on backdrop.
+            offerDetailLayover.addEventListener("click", (e) => {
+                if (e.target === offerDetailLayover) {
+                    offerDetailLayover.classList.toggle("hidden");
+                }
+            });
+        }
```
charludo, line 26: Awesome 🎉

charludo (changes requested):
Very nice!

I am experiencing the following issue though: In the applications list, the uppermost offer opens the "details" overlay when clicking the "details" button. For all other offers, nothing happens when pressing the button.
The reason is a mismatch between button and overlay IDs: in the `vote_form.html` template, you use the `offer_version.id` for the id of the details buttons, in `offer-detail-overlay.html` you use the `offer.id` as the id.
(And funnily enough the first pending offer in the view is version 1 of offer 1 😄)

JoeyStk:
> The reason is a mismatch between button and overlay IDs: in the vote_form.html template, you use the offer_version.id for the id of the details buttons, in offer-detail-overlay.html you use the offer.id as the id.

Ohh crap. Yes, this is a remaining issue from what I had tried earlier. Good catch! It should be fixed now. Now I'm matching the offers to each other :sweat_smile: 

Also this PR is blocked until https://github.com/digitalfabrik/integreat-compass/pull/105 is merged, because for reports we also have a details button that should open the overlay

charludo:
> Also this PR is blocked until #105 is merged, because for reports we also have a details button that should open the overlay

Also if you are already on this template topic, would you mind adding the title image to the vote forms as well? 😅 Just saw that in the initial designs it's included there

JoeyStk:
@charludo This PR is now ready for review with all discussed extras :grin:

charludo (approved):
Thank you! Works perfectly now

PR #123, opened 2023-09-04
Add modal for declining / accepting offers

### LINKED ISSUE #58: Modal for entering reason for declining a offer
### Motivation
A board member should enter a reason, why a offer is declined.

### Proposed Solution
See Figma screens and discussion on Figma: 

decline

JoeyStk:
### Short description

Add the overlay for declining and accepting an offer

### Proposed changes

- Add modal for accepting or declining an offer
- Handle reason why the offer was declined and store it in database

### Side effects

-  I think if when we add the next modal it would be a good idea to add like an abstract modal template which the modals can inherit from. However this would take a bit effort and therefore I think it's better if we do it after September 15th

### Resolved issues

Fixes: #58


charludo (commented):
adding `"reason": "reason",` to this line: https://github.com/digitalfabrik/integreat-compass/blob/4a807e6f2f034f99e7f38cf3c16fcf4824d196fc/tests/cms/views/interactions/test_vote_action.py#L11 should fix the tests

integreat_compass/static/src/js/application-overlay.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+};
+
+const closeOnBackdrop = (modal: HTMLElement) => {
+    modal.addEventListener("click", (e) => {
+        if (e.target === modal) {
+            modal.classList.toggle("hidden");
+        }
+    });
+};
```
charludo: The top two can definetly be combined, maybe even the third as well?

integreat_compass/static/src/js/application-overlay.ts
```diff
@@ -0,0 +1,52 @@
...
+            closeModal(modal, cancelButton);
+        }
+        if (modal && declineButton) {
+            openModal(modal, declineButton as HTMLElement);
+        }
+        if (modal) {
+            closeOnBackdrop(modal);
+        }
+    }
+};
```
charludo: This is also not super DRY - could you extract those to a function?

charludo (changes requested):
Nice! Apart from the below nitpicks, it is currently possible to submit rejections without a reason. I think we need both front- and backend checks for that.

charludo (approved):
🎉 Thank you! (And sorry for the merge conflicts 🙈)

PR #132, opened 2023-09-12
Fix reset button for map filter & adjust zoom on map

### LINKED ISSUE #129: Map in maximal zoom
When entering an address, map points at maximal zoom at that address, so the map is more or less blank.
It would be nice to choose a medium zoom level, so one can see the environment of that address

### LINKED ISSUE #130: Address is not resetted
### Describe the Bug
"zurücksetzen"-Button works only when a distance is selected.

### Steps to Reproduce

1. Add an address "Augsburg", press enter
2. klick on "zurücksetzen" -> address still present
3. Klick a distance and klick on "zurücksetzen" -> address and distance vanish

Address shoud be claered even without a distance

JoeyStk:
### Short description
This PR fixes the reset button for map filter, if there was no given distance. Also it adjusts the zoom on the map if there there is no radius set.

### Proposed changes

- Set default radius to 1. With the already existing implementation this already fixes both issues
- Add check if the reset-checkbox exists to avoid JS error message and failure of subsequent code

### Side effects

- Admittedly this is probably just a quick & dirty solution. If you have ideas for improvement, please let me know
- The current implementation doesn't activate the filter. In my experience this is the common way to use filters like this. Optionally we could also activate the filter with the default value of radius=1, which probably would result in no results most of the time or set the default value to radius=5 including a visual indication

### Resolved issues

Fixes: #129, 
Fixes: #130

### DIFF
--- a/integreat_compass/cms/templates/index.html
+++ b/integreat_compass/cms/templates/index.html
@@ -121,6 +121,7 @@
                                                class="filter-element hidden peer"
                                                type="radio"
                                                name="radius"
+                                               {% if distance == 5 %}checked{% endif %}
                                                {% if distance|slugify == filtered_radius %}checked{% endif %}
                                                value="{{ distance }}">
                                         <label for="distance-{{ distance }}"
--- a/integreat_compass/static/src/js/map.ts
+++ b/integreat_compass/static/src/js/map.ts
@@ -165,7 +165,6 @@ const initializeFilterMap = () => {
     initializeMap(container, searchBar, radiusToZoom(radius));
 
     reset.addEventListener("click", () => {
-        console.log("here");
         params.delete("radius");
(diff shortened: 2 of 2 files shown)


integreat_compass/cms/templates/index.html
```diff
@@ -121,6 +121,7 @@ <h2 class="text-gray-700 text-2xl font-semibold">{% translate "Filter" %}</h2>
                                                class="filter-element hidden peer"
                                                type="radio"
                                                name="radius"
+                                               {% if distance == 5 %}checked{% endif %}
                                                {% if distance|slugify == filtered_radius %}checked{% endif %}
```
charludo, line 125: That is WAY easier than I was thinking. Nice!

charludo (approved):
Thank you!!

PR #162, opened 2023-09-21
Fix overlay in report details not opening

### LINKED ISSUE #157: Popup window "Detail" not closable when even counted remarcs existent
### Describe the Bug
As a board member on "Meldungen" page the detail modal cannot be closed with the close button when the number of remarcs for one offer is an even number.  

### Steps to Reproduce

1. login as board member
2. goto "Meldungen" page
3. open detail page for an even numbered offer (Deutsch für Flüchtlinge ...)
4. try to close the popup

JoeyStk:
### Short description

This PR is supposed to fix the issue that the overlay in the report list is not open & closing properly. While inspecting this issue I found out that the issue stemmed from a mix-up of ids. For every offer there was one overlay linked to it. However with reports it's possible to create more than one reports for the same offer. This means that the offer id is not unique anymore and therefore we had the issues with opening and closing the (right) overlay. 

### Proposed changes

- Change from offer id to report id as unique identifier in report list
- Change from offer id to offer_version id as unique identifier in vote form (please take some time to really think about this and if there might be some unwanted side effects with this, that I might not be thinking of).

### Side effects

- Not unlikely, I couldn't didn't come up with one thought. Maybe you can think of some?

### Resolved issues

Fixes: #157

### DIFF
--- a/integreat_compass/cms/templates/index.html
+++ b/integreat_compass/cms/templates/index.html
@@ -173,7 +173,7 @@
                                     <p class="text-gray-700">{% translate "Created at:" %} {{ offer.public_version.created_at|date:"d. M Y" }}</p>
                                 </div>
                             </div>
-                            {% include "overlay/offer_detail_overlay.html" with public_version=True offer=offer offer_version=offer.public_version %}
+                            {% include "overlay/offer_detail_overlay.html" with public_version=True id=offer.id offer=offer offer_version=offer.public_version %}
                         {% endfor %}
                     {% else %}
                         {% translate "No offers could be found" %}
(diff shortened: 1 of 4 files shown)


charludo (approved):
If I'm understanding this correctly, we're including the details overlay once per report, and really opening "different" overlays for each report on the same offer, correct?

I don't think this is necessarily ideal, but neither is it a problem, so I'd say let's go with your solution, since it fixes the bug! 😄🎉 

Long term we should probably group the reports by offers anyways, in which case both the bug and the duplication issues become irrelevant, but that's something for the backlog.


About the `vote_form`, the change only affects the overlay, right? Not the vote submission? In which case I'd say there's no danger :D

JoeyStk:
> If I'm understanding this correctly, we're including the details overlay once per report, and really opening "different" overlays for each report on the same offer, correct?

Yep, exactly 


> About the vote_form, the change only affects the overlay, right? Not the vote submission? In which case I'd say there's no danger :D

Exactly

PR #174, opened 2026-07-01
Make filter chips full height

svenseeberg:
Fix not full height filter chips:

image


JoeyStk (approved):
LGTM!

svenseeberg:
> LGTM!

Sadly it does not yet yield the expected result :O

